PR #6, opened 2020-09-24
Add basic BuildKit tests in GitHub Actions

### LINKED ISSUE #1: DOCKER_BUILDKIT not working
I understand this docker version has BUILDKIT. To be enabled, in no snapped docker there are two options:
`$ DOCKER_BUILDKIT=1 docker build .`

But, trying docker docs simple example (which works fine with not snapped docker), with docker snap I obtain **parse** errors. (Doesn't recognize any buildkit characteristic in RUN statements). So, I wanted to use the second option to use this characteristic:

```
To enable docker BuildKit by default, set daemon configuration in /etc/docker/daemon.json feature to true and restart the daemon:

{ "features": { "buildkit": true } }
```

This config file is the repo, but once is installed I can't modify it. 

Any workaround?

tianon:
This is an attempt to validate #1 (although this is testing the newly updated 19.03.13 from #4, so not a *completely* fair test for that issue, but at least gives us a clue if there's light at the end of that tunnel).

**Edit:** with the backported patch from 19.03.15 (pending #13), this actually fixes #1 :partying_face:

### DIFF
--- a/dockerd-patches/buildkit-apparmor.patch
+++ b/dockerd-patches/buildkit-apparmor.patch
@@ -1,0 +1,291 @@
+From 5472f39022e99c14b2f055eac4d9619e3663ae20 Mon Sep 17 00:00:00 2001
+From: Brian Goff <cpuguy83@gmail.com>
+Date: Fri, 9 Oct 2020 17:20:48 +0000
+Subject: [PATCH] buildkit: Apply apparmor profile
+
+Signed-off-by: Brian Goff <cpuguy83@gmail.com>
+(cherry picked from commit 611eb6ffb32aa37876b4b47cec12e4fd47610838)
+
+Renamed constant defaultAppArmorProfile to defaultApparmorProfile.
+
+Signed-off-by: Tibor Vass <tibor@docker.com>
+---
+ builder/builder-next/builder.go          | 1 +
+ builder/builder-next/controller.go       | 2 +-
+ builder/builder-next/executor_unix.go    | 3 ++-
+ builder/builder-next/executor_windows.go | 2 +-
+ cmd/dockerd/daemon.go                    | 1 +
+ daemon/apparmor_default.go               | 8 ++++++++
+ daemon/apparmor_default_unsupported.go   | 5 +++++
+ 7 files changed, 19 insertions(+), 3 deletions(-)
+
+diff --git a/builder/builder-next/builder.go b/builder/builder-next/builder.go
+index 10c0083a31c2..f334a39267bc 100644
+--- a/builder/builder-next/builder.go
++++ b/builder/builder-next/builder.go
+@@ -76,6 +76,7 @@ type Opt struct {
+ 	Rootless            bool
+ 	IdentityMapping     *idtools.IdentityMapping
+ 	DNSConfig           config.DNSConfig
++	ApparmorProfile     string
+ }
+ 
+ // Builder can build using BuildKit backend
+diff --git a/builder/builder-next/controller.go b/builder/builder-next/controller.go
+index 33d3f549642e..d5fa7f6c290b 100644
+--- a/builder/builder-next/controller.go
++++ b/builder/builder-next/controller.go
+@@ -116,7 +116,7 @@ func newController(rt http.RoundTripper, opt Opt) (*control.Controller, error) {
+ 
+ 	dns := getDNSConfig(opt.DNSConfig)
+ 
+-	exec, err := newExecutor(root, opt.DefaultCgroupParent, opt.NetworkController, dns, opt.Rootless, opt.IdentityMapping)
++	exec, err := newExecutor(root, opt.DefaultCgroupParent, opt.NetworkController, dns, opt.Rootless, opt.IdentityMapping, opt.ApparmorProfile)
+ 	if err != nil {
+ 		return nil, err
+ 	}
+diff --git a/builder/builder-next/executor_unix.go b/builder/builder-next/executor_unix.go
+index d684b9f6e2e6..d04334a0046e 100644
+--- a/builder/builder-next/executor_unix.go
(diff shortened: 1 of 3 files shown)


tianon:
Trying to build this locally to see if whatever additional things GitHub Actions is doing is the cause of this (like it was for so much else while I was working on #4), or whether it's something deeper like a new AppArmor denial (I was able to do at least the basic `debian-hello` test with BuildKit on a 19.03.11 build I had installed locally).

tianon:
:disappointed: 

```console
root@snappy:~# snap list docker
Name    Version   Rev  Tracking     Publisher   Notes
docker  19.03.13  516  latest/edge  canonical✓  -

root@snappy:~# DOCKER_BUILDKIT=1 docker build --pull -t debian-hello - <<'EOF'
FROM debian:buster-slim
RUN set -eux; \
    apt-get update; \
    apt-get install -y hello; \
    rm -rf /var/lib/apt/lists/*
CMD ["hello"]
EOF
[+] Building 2.6s (6/6) FINISHED                                                                                                                                      
 => [internal] load build definition from Dockerfile                                                                                                             0.1s
 => => transferring dockerfile: 179B                                                                                                                             0.0s
 => [internal] load .dockerignore                                                                                                                                0.1s
 => => transferring context: 2B                                                                                                                                  0.0s
 => [internal] load metadata for docker.io/library/debian:buster-slim                                                                                            2.3s
 => [1/2] FROM docker.io/library/debian:buster-slim@sha256:b527bec3708ee1fe2ffb4625d742a60bb45b06064e7c64a81b16550fe42710f5                                      0.0s
 => CACHED [2/2] RUN set -eux;     apt-get update;     apt-get install -y hello;     rm -rf /var/lib/apt/lists/*                                                 0.0s
 => exporting to image                                                                                                                                           0.1s
 => => exporting layers                                                                                                                                          0.0s
 => => writing image sha256:f69f7451890e9e02f226bddb6771140990f9636e2d1080317654c9d2eb360311                                                                     0.0s
 => => naming to docker.io/library/debian-hello                                                                                                                  0.0s
```

tianon:
@anonymouse64 any ideas for what to add to try and debug this further?  I don't see anything obvious in `snap logs docker.dockerd`, `dmesg`, or any of `/var/log/*.log` for why it's getting `Permission denied` trying to run `apt-get`, but only from within BuildKit (especially since it works fine when I test it locally in what should be essentially the same environment -- Ubuntu 16.04, etc) :disappointed:

anonymouse64:
@tianon the first thing I see in your workflow is that you are not connecting the `network-control` interface, it's unlikely that's the cause of the buildkit issue but it's something. 

I will have a look, I'm not sure how buildkit works but maybe it's using something that is being denied. Also, I typically look via journalctl via `journalctl --no-pager | grep DENIED | grep docker` for denials, perhaps denials are not going to dmesg properly or something.

tianon:
Interesting:

```
Sep 28 18:23:20 fv-az138 audit[13412]: AVC apparmor="DENIED" operation="file_lock" profile="snap.docker.dockerd" name="/dev/null" pid=13412 comm="exe" requested_mask="k" denied_mask="k" fsuid=0 ouid=0
Sep 28 18:23:21 fv-az138 audit[13633]: AVC apparmor="DENIED" operation="file_lock" profile="snap.docker.dockerd" name="/dev/null" pid=13633 comm="exe" requested_mask="k" denied_mask="k" fsuid=0 ouid=0
Sep 28 18:23:21 fv-az138 audit[13649]: AVC apparmor="DENIED" operation="exec" profile="snap.docker.dockerd" name="/usr/bin/apt-get" pid=13649 comm="sh" requested_mask="x" denied_mask="x" fsuid=0 ouid=0
```

tianon:
I'm over my head for sure here, but it *seems* like it's using the wrong AppArmor profile for running the container in BuildKit?

anonymouse64:
> Sep 28 18:23:21 fv-az138 audit[13633]: AVC apparmor="DENIED" operation="file_lock" profile="snap.docker.dockerd" name="/dev/null" pid=13633 comm="exe" requested_mask="k" denied_mask="k" fsuid=0 ouid=0

no idea what dockerd is trying to do here :confused: but it's probably just noise

> Sep 28 18:23:21 fv-az138 audit[13649]: AVC apparmor="DENIED" operation="exec" profile="snap.docker.dockerd" name="/usr/bin/apt-get" pid=13649 comm="sh" requested_mask="x" denied_mask="x" fsuid=0 ouid=0

so here's the smoke from the smoking gun, this indicates that for whatever reason dockerd did not transition to the apparmor profile for the container before trying to run this, unclear why, but maybe buildkit doesn't use confinement for some reason when launching containers? Do you have any pointers to the implementation of buildkit or more thorough documentation on what the difference is internally between legacy builds and buildkit builds?

tianon:
I think we might've exposed an upstream bug here (where BuildKit might not be using the correct AppArmor profile for the containers it spawns) -- there are some folks on the Docker side digging further. :+1:

dockerd-patches/buildkit-apparmor.patch
```diff
@@ -0,0 +1,291 @@
+From 5472f39022e99c14b2f055eac4d9619e3663ae20 Mon Sep 17 00:00:00 2001
```
tianon, line 1: Including this patch file is hopefully temporary pending #13 -- I'd much rather update to 19.03.15 than carry this patch manually but this is what's needed to make this testing actually possible. :sweat_smile:

tianon:
Split this into two commits so that we have cf4a338 which fixes #1 and d9f67a4 which tests/validates it via GitHub Actions (with the minor change necessary for `snap/snapcraft.yml` - so we can include a clean revert of d9f67a4 in #13).

PR #9, opened 2020-10-08
Update dockerd-wrapper

diddledan:
Use `/run/snap.$SNAP_INSTANCE_NAME` for `docker.pid` and `exec-root` so that an unclean shutdown of the system does not leave a stale pidfile or other runtime-only files preventing Docker from starting up once the system is rebooted.

See: https://forum.snapcraft.io/t/docker-fails-to-restart-after-power-failure/19105/13

### DIFF
--- a/bin/dockerd-wrapper
+++ b/bin/dockerd-wrapper
@@ -63,6 +63,6 @@ exec dockerd \
 	--group "$default_socket_group" \
 	--exec-root="$SNAP_DATA/run/docker" \
 	--data-root="$SNAP_COMMON/var-lib-docker" \
-	--pidfile="$SNAP_DATA/run/docker.pid" \
+	--pidfile="/tmp/docker.pid" \
 	--config-file="$SNAP_DATA/config/daemon.json" \
 	"$@"
--- a/bin/dockerd-wrapper
+++ b/bin/dockerd-wrapper
@@ -45,11 +45,12 @@ rm -rf "$dir"
 trap - EXIT
 
 # modify XDG_RUNTIME_DIR to be a snap writable dir underneath $SNAP_COMMON 
-# until LP #1656340 is fixed
+# (until https://bugs.launchpad.net/snappy/+bug/1656340 is fixed)
 export XDG_RUNTIME_DIR=$SNAP_COMMON/run
 mkdir -p "$XDG_RUNTIME_DIR"
 
-# use SNAP_DATA for most "data" bits
+# use SNAP_COMMON for "/var/lib/docker" since it's common across snap revisions
+# (and typically gets large, so we don't want it backed up and restored like SNAP_DATA)
 mkdir -p "$SNAP_COMMON/var-lib-docker"
 
 # use "/run/snap.$SNAP_INSTANCE_NAME" for ephemeral storage
--- a/bin/dockerd-wrapper
+++ b/bin/dockerd-wrapper
@@ -47,13 +47,16 @@ trap - EXIT
 # modify XDG_RUNTIME_DIR to be a snap writable dir underneath $SNAP_COMMON 
 # until LP #1656340 is fixed
 export XDG_RUNTIME_DIR=$SNAP_COMMON/run
+mkdir -p "$XDG_RUNTIME_DIR"
 
 # use SNAP_DATA for most "data" bits
-mkdir -p \
-	"$SNAP_DATA/run" \
-	"/run/snap.$SNAP_INSTANCE_NAME" \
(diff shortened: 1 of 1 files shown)


bin/dockerd-wrapper
```diff
@@ -51,7 +51,7 @@ export XDG_RUNTIME_DIR=$SNAP_COMMON/run
 # use SNAP_DATA for most "data" bits
 mkdir -p \
 	"$SNAP_DATA/run" \
-	"$SNAP_DATA/run/docker" \
+	"/run/snap.$SNAP_INSTANCE_NAME" \
```
tianon, line 54: Ah, interesting!  Is this an official path that'd documented somewhere for our use?
(Shouldn't there be another variable like `$SNAP_DATA` to point to it?  I don't see anything related to `/run` in https://snapcraft.io/docs/environment-variables :sweat_smile:)

If we switch to this, I don't think we need to create `"$SNAP_DATA/run"` at all, since `--exec-root` and `--pidfile` are all we use it for AFAIR, and it's much better (matches upstream's expectations) to put all that on a tmpfs. :smile:

diddledan: @anonymouse64 I don't recall seeing this documented, either.. I've tested the feature works, so I think the docs just haven't caught up with this feature yet...

tianon: I see it defined in the common AppArmor profile at https://github.com/snapcore/snapd/blob/79aced6888fae0a5785c784d4aa4dba28992c236/interfaces/apparmor/template.go#L444-L447, so perhaps that's the closest thing we have to official "documentation" of it?

tianon: (also in the packaging changelog at https://github.com/snapcore/snapd/blob/79aced6888fae0a5785c784d4aa4dba28992c236/packaging/ubuntu-16.04/changelog#L3766, but that's probably just sourced from the commit/PR title)

anonymouse64: It's not documented, but it should be :wink: , and there's not an env variable, but there should be :wink: 

see: https://github.com/snapcore/snapd/blob/79aced6888fae0a5785c784d4aa4dba28992c236/interfaces/apparmor/template.go#L444-L447

diddledan: @degville, we need this in a document please :-)

degville: > @degville, we need this in a document please :-)

will do! thanks for letting me know.

tianon: So, I'd like to adjust our implementation of it here a little bit for now so we have a better chance of remembering. :sweat_smile:

Here's what I've got in mind:

```bash
# modify XDG_RUNTIME_DIR to be a snap writable dir underneath $SNAP_COMMON 
# (until https://bugs.launchpad.net/snappy/+bug/1656340 is fixed)
export XDG_RUNTIME_DIR="$SNAP_COMMON/run"
mkdir -p "$XDG_RUNTIME_DIR"

# use SNAP_COMMON for "/var/lib/docker" since it's common across snap revisions
# (and typically gets large, so we don't want it backed up and restored like SNAP_DATA)
mkdir -p "$SNAP_COMMON/var-lib-docker"

# use "/run/snap.$SNAP_INSTANCE_NAME" for ephemeral storage
# - https://github.com/snapcore/snapd/pull/6109
# - https://github.com/docker-snap/docker-snap/pull/9
run_dir="/run/snap.$SNAP_INSTANCE_NAME"
mkdir -p "$run_dir"

workaround_lp1606510

workaround_apparmor_profile_reload

exec dockerd \
	--group "$default_socket_group" \
	--exec-root="$run_dir" \
	--data-root="$SNAP_COMMON/var-lib-docker" \
	--pidfile="$run_dir/docker.pid" \
	--config-file="$SNAP_DATA/config/daemon.json" \
	"$@"
```

(Happy to take over and update/carry this PR directly if you'd prefer :+1:)

tianon: @degville I'm by no means any kind of authority (just a humble nobody here :sweat_smile:), but it might also be worth adding a note about `XDG_RUNTIME_DIR` (which is set) that includes a caveat about https://bugs.launchpad.net/snappy/+bug/1656340 (which makes the provided value harder to use) :sweat_smile:

diddledan: @tianon, I've added 8cc8d9c to include your suggestions

tianon:
Looks good! Thanks for the PR, the conversation, and for seeing this through. 👍

PR #11, opened 2020-10-31
Add docker group instructions and certs info to description

galgalesh:
This PR

* adds additional info to the description. This info was already available in the README but not in the description.
* removes the line about `docker.compose` from the readme (there seems to be an alias in place)
* fixes some linter errors in the readme

Note: my previous changes to the description did not propagate to the store page. This is probably because the description in the store was manually changed. You might need to run `snapcraft upload-metadata` explicitly before the metadata in the store will be changed.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -5,8 +5,25 @@ summary: Docker container runtime
 description: |
   Build and run container images with Docker.
 
+  **Usage**
+
   * This build can only access files in the home directory. So Dockerfiles and all other files used in commands like `docker build`, `docker save` and `docker load` need to be in $HOME.
   * You can change the configuration of this build by modifying the files in `/var/snap/docker/current/`.
+  * Additional certificates used by the Docker daemon to authenticate with registries need to be added in `/var/snap/docker/common/etc/certs.d` (instead of `/etc/docker/certs.d`).
+
+  **Running Docker as normal user**
+
+  By default, Docker is only accessible with root privileges (`sudo`). If you want to use docker as a regular user, you need to add your user to the `docker` group.
+
+      sudo addgroup --system docker
+      sudo adduser $USER docker
+      newgrp docker
+      sudo snap disable docker
+      sudo snap enable docker
+
+  **Warning:** if you add your user to the `docker` group, it will have similar power as the `root` user. For details on how this impacts security in your system, see https://docs.docker.com/engine/security/#docker-daemon-attack-surface
+
+  **Authors**
 
   This snap is built by Canonical based on source code published by Docker, Inc. It is not endorsed or published by Docker, Inc.
 
--- a/README.md
+++ b/README.md
@@ -8,33 +8,33 @@ This repository contains the source for the `docker` snap package.  The package 
 
 To install the latest stable release of Docker CE using `snap`:
 
-```
-$ sudo snap install docker
+```shell
+sudo snap install docker
 ```
 
 If you are using Ubuntu Core 16,
 
 * Connect the `docker:home` plug as it's not auto-connected by default:
 
-```
-$ sudo snap connect docker:home
+```shell
+sudo snap connect docker:home
 ```
 
 If you are using an alternative snap-compatible Linux distribution ("classic" in snap lingo), and would like to run `docker` as a normal user:
 
 * Create and join the `docker` group.
 
-```
-$ sudo addgroup --system docker
-$ sudo adduser $USER docker
-$ newgrp docker
+```shell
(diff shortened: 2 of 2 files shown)


tianon:
Seems generally sane to me!  Thank you.

> Note: my previous changes to the description did not propagate to the store page. This is probably because the description in the store was manually changed. You might need to run `snapcraft upload-metadata` explicitly before the metadata in the store will be changed.

This is where things get dicey for me -- I don't actually use `snapcraft` locally for anything except performing test builds.  The builds on https://snapcraft.io/docker get built by the (semi-new?) post-commit automation there.  Also, I don't know whether a build with your changes has actually propagated to the stable channel, so perhaps that's why the description hasn't auto-updated?  I would've expected the snapcraft.io auto-builds feature to be more intelligent about this, is there perhaps a way to "reset" the metadata and request that it be regenerated from the snap metadata instead?

galgalesh:
> Also, I don't know whether a build with your changes has actually propagated to the stable channel, so perhaps that's why the description hasn't auto-updated?

In the past, updates to the edge channel also change the stable snap metadata in the store in the past. I filed an issue for this, but the devs decided this was expected behavior.

> is there perhaps a way to "reset" the metadata and request that it be regenerated from the snap metadata instead?

As far as I know, the only way to reset the metadata is to force-update the metadata using the snapcraft tool. Once you update it from the store GUI, it won't get automatically updated anymore until you force-update it using the cli.

snap/snapcraft.yaml
```diff
@@ -5,8 +5,25 @@ summary: Docker container runtime
...
+
+  By default, Docker is only accessible with root privileges (`sudo`). If you want to use docker as a regular user, you need to add your user to the `docker` group.
+
+      sudo addgroup --system docker
+      sudo adduser $USER docker
+      newgrp docker
+      sudo snap disable docker
+      sudo snap enable docker
+
+  **Warning:** if you add your user to the `docker` group, it will have similar power as the `root` user. For details on how this impacts security in your system, see https://docs.docker.com/engine/security/#docker-daemon-attack-surface
```
tianon, line 24: Before we add this to the official documentation (and thus encourage it to some degree): @anonymouse64, wasn't there an issue around the `docker` group and potential problems?  Do you think we should hold off documenting this until that's resolved in a satisfactory way such that the `docker` group is deemed reasonable to rely on?

tianon:
> > Also, I don't know whether a build with your changes has actually propagated to the stable channel, so perhaps that's why the description hasn't auto-updated?
> 
> In the past, updates to the edge channel also change the stable snap metadata in the store in the past. I filed an issue for this, but the devs decided this was expected behavior.
> 
> > is there perhaps a way to "reset" the metadata and request that it be regenerated from the snap metadata instead?
> 
> As far as I know, the only way to reset the metadata is to force-update the metadata using the snapcraft tool. Once you update it from the store GUI, it won't get automatically updated anymore until you force-update it using the cli.

Ah, fair enough, I suppose (seems like somewhere the "Builds" feature could be improved :sweat_smile:).

I'll see what I can do with the local client to get the current merged changes up so it can hopefully auto-sync when we get these changes in. :+1:

snap/snapcraft.yaml
anonymouse64: I think it's fine to document, it's unlikely that there will be any changes to how that works from a user perspective, and the new users who do this are dwarfed by the many many existing installs that already are doing this so it's not like we will be introducing a large group of new affected users by this

tianon:
... I think I managed to Do The Thing: https://snapcraft.io/docker :innocent:

(Gonna merge this now)

tianon:
Doh, builds are failing thanks to what I hacked around in #23 ...
(underlying fix in https://github.com/snapcore/snapcraft/pull/3429 is not in a release yet :weary:)

anonymouse64:
perhaps is reason enough to upgrade to base: core18 or core20 :open_mouth:  ? https://github.com/docker-snap/docker-snap/issues/15

tianon:
I'm not running snap, lxd, or multipass locally (and none of them work super great inside a Docker container), and my dev environment is a VM that doesn't do nested VMs, so the workflow changes required for that are kind of a large hurdle for me (and the biggest reason that's not resolved beyond the issues listed in #15 like what to do about i386). :confused:

galgalesh:
In case it helps; if you're using `base: core18`, you can direct snapcraft to use an LXD container instead of a VM by using `snapcraft --use-lxd`. This is useful in VM's.

More info: https://snapcraft.io/docs/build-on-lxd

PR #14, opened 2020-12-08
Upload an example "docker.snap" for testing on PRs

tianon:
:eyes:

(https://github.com/docker-snap/docker-snap/pull/13#issuecomment-740232256)

PR #17, opened 2020-12-29
Add "libsystemd-dev" so the journald log driver can read

### LINKED ISSUE #16: Impossible to use docker logs (at least when using journald)
After some small modifications to home-assistant's supervised-installer (see https://github.com/home-assistant/supervised-installer/pull/140) I've been able to correctly install and use hassio in my raspberry pi4 running Ubuntu with docker-snap.
Once it's installed in a snap-readable prefix all works fine, but unfortunately it seems that logs are not properly readable via `docker logs homeassistant` (even though I can read them via journalctl):

```
Dec 29 00:57:33 uberry docker.dockerd[1760]: time="2020-12-29T00:57:33.891258733+01:00" level=error msg="Handler for GET /v1.40/containers/23984b41e39fbb180ce14baf502c5e86975216cde6ca0b481ddf863c4a05a63f/logs returned error: configured logging driver does not support reading"
```

To be clear, I've modified (before installing hassio) my `$(snap run --shell docker -c 'printenv SNAP_DATA')/config/daemon.json` to use:

```json
{
    "log-level":        "error",
    "storage-driver":   "overlay2",
    "log-driver":       "journald"
}
```

tianon:
Fixes #16

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -81,6 +81,18 @@ jobs:
       - name: Hello World
         run: sudo docker run --rm hello-world
 
+      - name: Hello World (journald)
+        run: |
+          trap 'sudo dmesg | tail -n20; sudo snap logs -n=20 docker.dockerd' ERR
+          expectedOutput="testing-journald-log-driver-$RANDOM-$RANDOM"
+          sudo docker run --name test-journald --log-driver journald bash -c 'echo "$@"' -- "$expectedOutput"
+          actualOutput="$(sudo docker logs test-journald)"
+          [ -z "$actualOutput" ] # verify that "log-observe" is still necessary
+          sudo snap connect docker:log-observe
+          actualOutput="$(sudo docker logs test-journald)"
+          [ "$actualOutput" = "$expectedOutput" ]
+          docker rm test-journald
+
       - name: Docker Build
         run: |
           sudo docker build --pull -t debian-hello - <<'EOF'
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -64,6 +64,7 @@ apps:
     plugs:
       - firewall-control
       - home
+      - log-observe
       - network-bind
       - network-control
       - privileged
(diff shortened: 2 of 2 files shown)


3v1n0:
Is failure due to missing log-observe interface or something else?

tianon:
Good question!  I'm going to add a `trap` so we can (hopefully) get some logs from the daemon and see what's going on.

tianon:
Yep, turns out `log-observe` *is* also necessary for this to work (so even after this is merged/released, you'll have to `snap connect docker:log-observe` to let Docker read the journal logs back).

3v1n0:
Well doesn't seem a too big deal, maybe could be even auto-connected one day, but in any case it's fine IMHO.

tianon:
FWIW, if you wanted to test this sooner (since there are a few other things baking currently which will prevent this from being merged/released in the short term), you can download a pre-built version from https://github.com/docker-snap/docker-snap/suites/1740189892/artifacts/33019489 (thanks to #14) and install it with something like `snap install --dangerous docker.snap`

anonymouse64 (approved):
lgtm, but please request an auto-connect on the forum for log-observe, I see no reason why docker shouldn't get an auto-connect for this seeing as how it already has docker-support auto-connected.

tianon:
I'm not convinced it makes sense to auto-connect?  There's no default-enabled core functionality which requires/uses it, and it's a potential information leak, so it seems more "correct" (and in line with the spirit of the explicit `privileged` interface) to leave it disconnected and ask users who want/need it to connect it IMO.

anonymouse64:
@tianon Ah, I guess I misunderstood the issue, I assumed that the issue was that _any_ `docker logs` command didn't work, is it just that `docker logs` doesn't work when you configure this specific `logging-driver`? 

In any case, there is already loads of things that the docker snap is allowed via the `docker-support` interface so I don't think it's really in practice an issue to have this additional vulnerability leak and it would make it easier for users to use the snap fully in place of the deb, but that's just my two sense.

tianon:
Yes, this _only_ affects users who are using `--log-driver=systemd` or have specifically modified their `daemon.json` configuration to use the `systemd` log driver by default.

I imagine some of the other log drivers might have issues too (or some of the esoteric options to the default `json-file` log driver), but I suppose most of them are "just" network-based so will probably work fine?

tianon:
Regression potential here is extremely low, and IMO having this auto-connect is a separate thread (having it available as an optional connection gives us some concrete functionality we didn't have before).

Also, to be clear, we can _write_ to the logs without the interface connected -- it's only reading them back via `docker logs` that doesn't work (reading them via `journalctl` is unaffected).

PR #26, opened 2021-02-04
Update to "base: core18"

### LINKED ISSUE #15: Upgrade to base: core20
(copying salient points from https://github.com/docker-snap/docker-snap/issues/8 to a dedicated issue)

> There also would be the potentially significant amount of snapcraft'ing involved in porting the snapcraft.yaml to use `base: core20` instead of our current implicit `core` base.

> Yeah, porting the Snap to `base: core20` has been on my TODO list for a while to at least consider and look into deeper, but setting up a development environment for building `base: core20` snaps is _way_ more complicated than `base: core`, so it's something that's hard for me to be motivated to spend a lot of time on.

> Porting to core20 also brings with it the complicated question of supporting i386 - do we drop it or do we maintain an i386 only version of the snap based on core16/core18? See https://forum.snapcraft.io/t/how-to-best-handle-i386-when-moving-a-snap-to-base-core20/17680 for more discussion

tianon:
Fixes #15

This will help us resolve things like https://github.com/snapcore/snapcraft/pull/3429 (which is currently blocking our ability to update the snap at all...) and hopefully let us use `stage-packages: git` instead of building our own bespoke `git` from source. :eyes:

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -28,17 +28,16 @@ description: |
   This snap is built by Canonical based on source code published by Docker, Inc. It is not endorsed or published by Docker, Inc.
 
   Docker and the Docker logo are trademarks or registered trademarks of Docker, Inc. in the United States and/or other countries. Docker, Inc. and other parties may also have trademark rights in other terms used herein.
-
-confinement: strict
+license: (Apache-2.0 AND MIT AND GPL-2.0)
 grade: stable
 
+base: core18
+confinement: strict
 assumes: [snapd2.40]
 
-passthrough:
-  license: (MIT AND Apache-2.0 AND GPL-2.0)
-  layout:
-    /etc/docker:
-      bind: $SNAP_DATA/etc/docker
+layout:
+  /etc/docker:
+    bind: $SNAP_DATA/etc/docker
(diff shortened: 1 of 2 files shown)


tianon:
Success!

snap/snapcraft.yaml
```diff
@@ -28,14 +28,14 @@ description: |
...
 
+base: core18
+confinement: strict
 assumes: [snapd2.40]
 
 passthrough:
-  license: (MIT AND Apache-2.0 AND GPL-2.0)
   layout:
     /etc/docker:
       bind: $SNAP_DATA/etc/docker
```
anonymouse64, line 41: FWIW, this can come out of the `passthrough` too

anonymouse64 (approved):
very nice!

snap/snapcraft.yaml
tianon: Ah, I thought it might, but it wasn't included in https://snapcraft.io/docs/snapcraft-yaml-reference so I figured better safe than sorry. :smile:

anonymouse64: I asked to get the docs updated to mention that

PR #27, opened 2021-02-06
Stop compiling Git from source

tianon:
Now that we're on "core18", this should give us a new enough Git that we don't need our own version anymore.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -37,7 +37,13 @@ assumes: [snapd2.48.2] # TODO bump this to 2.49 once that's stable (https://gith
 
 layout:
   /etc/docker:
-    bind: $SNAP_DATA/etc/docker
+    symlink: $SNAP_DATA/etc/docker
+  /etc/gitconfig:
+    bind-file: $SNAP_DATA/etc/gitconfig
+  /usr/lib/git-core:
+    symlink: $SNAP/usr/lib/git-core
+  /usr/share/git-core/templates:
+    symlink: $SNAP/usr/share/git-core/templates
 
 plugs:
   home:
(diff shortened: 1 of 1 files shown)


tianon:
Hahahahahahahaha :sob: 

````
unable to prepare context: unable to 'git clone' to temporary context directory: failed to init repo at /tmp/docker-build-git182722922: warning: unable to access '/etc/gitconfig': Permission denied
fatal: unknown error occurred while reading the configuration files
: exit status 128
```

tianon:
IT WORKS!!!

snap/snapcraft.yaml
```diff
@@ -37,7 +37,13 @@ assumes: [snapd2.48.2] # TODO bump this to 2.49 once that's stable (https://gith
 
 layout:
   /etc/docker:
-    bind: $SNAP_DATA/etc/docker
+    symlink: $SNAP_DATA/etc/docker
+  /etc/gitconfig:
+    bind-file: $SNAP_DATA/etc/gitconfig
```
anonymouse64, line 42: nice!

anonymouse64 (approved):
nicely done

snap/snapcraft.yaml
tianon: Indeed!  I was reading about `layout:` (esp. the warning about `bind:` having performance impacts that I frankly don't understand completely), and tried to use `symlink:` here instead, but it then fails during `snap install` if the host already has `/etc/gitconfig` :sweat_smile:

I was also afraid `bind-file:` would be sad that the RHS doesn't exist, but it doesn't mind one bit. :smile:

tianon: (and that's why I changed `/etc/docker` to be a symlink instead -- hopefully we don't see that breaking folks in a similar way to what I saw with `/etc/gitconfig` :eyes: :grimacing:)

anonymouse64: I for some reason never responded to this, but to be clear, the rules for layouts are that the right hand side must satisfy one of the following:

1. doesn't exist at all
2. exist, but be of the same file type as the type declared in the layout

The second rule means that if you are using a `symlink: ...`, the RHS always needs to be a symlink on the user's machine (or not exist), or if you are using `bind: ...` then the RHS always needs to be a directory (or not exist) or if you are using `bind-file: ...` then the RHS needs to be a file (or not exist).

Hopefully that brings some clarity, and re: performance it's super negligible for a snap like docker, it's more noticable (but still like less than a second the first time and is cached) when someone opens a graphical snap that has a lot of these layouts and also uses content interfaces, etc.

PR #29, opened 2021-02-22
WIP hack to test build on i386

tianon:
(If I can figure out a way to make this successful, it'll turn into a proper build matrix so we can test both.)

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -11,7 +11,7 @@ defaults:
 jobs:
   smoke-test:
     name: Snapcraft
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


tianon:
Heh, it fails, probably because we're missing some of the setup steps from https://github.com/snapcore/snapcraft/blob/4ac0faaf436a6d3e31fe06cd154c552b32b9bffd/snapcraft/internal/build_providers/_lxd/_lxd.py#L412-L470

(`snapcraft` doesn't run them unless it creates the container itself, which makes sense since they only need to happen once)

tianon:
I managed to use one of the i386 images from http://cloud-images.ubuntu.com/bionic/current/ to test the Python `cryptography` issue locally (which is what I was trying to get this working to test in an automated way), so I don't need this quite as urgently, but I'd still love to have a solution (so I'm going to leave this open as a draft).

PR #32, opened 2021-03-10
Update to Git v2.26.3 (for CVE-2021-21300)

tianon:


tianon:
(I don't think this CVE has a high chance of being exploited with how we run `git`, but they updated the specific minor release we support, so we might as well upgrade. :+1:)

PR #41, opened 2021-06-07
Update to 20.10.7

### LINKED ISSUE #40: sysctl command not allowed in host network namespace: unknown
Hi there,

We run into an issue with docker on `latest/edge` channel. A sysctl command will run in the container using host network. Here is the API log from Docker:

```
msg="Handler for POST /containers/8483159eeddf0962ea101db60eb26b787a2498147ab494cb2da96c9b289bff9d/start returned error: OCI runtime create failed: sysctl xxxx not allowed in host network namespace: unknown"
```

This issue was resolved after we installed docker from `latest/stable` channel.

The current versions of Docker snap:
- latest/edge: 19.03.13
- latest/edge: 20.10.6

Thanks,
Hao

tianon:
Closes #40

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,6 +1,6 @@
 name: docker
 #title: Docker
-version: '20.10.6'
+version: '20.10.7'
 summary: Docker container runtime
 description: |
   Build and run container images with Docker.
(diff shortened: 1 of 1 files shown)


anonymouse64 (approved):
looks pretty safe to me

tianon:
At this point, if the GitHub Actions pass, we're probably in pretty OK shape :joy:

PR #42, opened 2021-08-04
Update "certs.d" paths

tianon:
This is technically a follow-up to 419af60dba5af78ff695217aaf55c8dd70c5592c, which removed the patch that enabled "/var/snap/docker/common/etc/certs.d"


anonymouse64 (approved):
lgtm thanks for fixing this

PR #46, opened 2021-09-02
Replace symlink with bind for /etc/docker

jkfran:
Fixes https://github.com/docker-snap/docker-snap/issues/45

We can't expect `/etc/docker` to be empty. This issue was introduced recently in https://github.com/docker-snap/docker-snap/pull/27/

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -37,7 +37,7 @@ assumes: [snapd2.50]
 
 layout:
   /etc/docker:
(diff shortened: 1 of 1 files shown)


carkod (approved):

tianon (approved):

jkfran:
@tianon thank you! when would it be on `latest/stable`? Could you let me know?

tianon:
Sorry for the delay -- it should be now!

PR #49, opened 2021-09-17
Update "snapcore/action-build" to v1.0.9

tianon:
(Especially to avoid failing on `snap refresh lxd`)

https://github.com/snapcore/action-build/pull/24 :eyes:

See also https://forum.snapcraft.io/t/unexpectedly-empty-response-from-the-server/23393/16?u=tianon

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Build
         # https://github.com/snapcore/action-build/releases
-        uses: snapcore/action-build@v1.0.7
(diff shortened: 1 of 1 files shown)


anonymouse64:
this can now go back to tracking v1 which was updated to point to v1.0.9 now

tianon:
@anonymouse64 I've always had it at a specific version -- I like less surprises (or rather, surprises at more explicit update moments) :sweat_smile:

anonymouse64:
ah fair enough tbh I didn't even look at the diff here to see you had it on 1.0.7 before /o\ but good to know it's fixed here anyways

PR #51, opened 2021-09-24
Add docker-registry-certificates content interface slot

alexclewontin:
Currently, the only way to add additional registry certificates is to manually place them in the $SNAP_DATA/etc/docker/certs.d directory. This is fine on systems where the user has an interactive shell with access to that directory, but not particularly feasible on an unattended system (i.e. Ubuntu Core). This commit adds a content interface, which allows another snap to access that directory to provision it with certificates.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -9,7 +9,7 @@ description: |
 
   * This build can only access files in the home directory. So Dockerfiles and all other files used in commands like `docker build`, `docker save` and `docker load` need to be in $HOME.
   * You can change the configuration of this build by modifying the files in `/var/snap/docker/current/`.
-  * Additional certificates used by the Docker daemon to authenticate with registries need to be added in `/var/snap/docker/current/etc/docker/certs.d` (instead of `/etc/docker/certs.d`).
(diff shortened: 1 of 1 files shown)


anonymouse64 (commented):
lgtm, would be great if you could share a test case as well

tianon:
I'm a little bit concerned that the name might have some overlap with a potential future `docker-registry` snap -- @anonymouse64 do you think that's warranted?  Should we adjust this name to make it more clear it's certificates for `docker-daemon`, not `docker-registry`?  (No strong opinion here -- if you think this is solid as-is, I'm good! :sweat_smile:)

anonymouse64:
@tianon There won't be any overlap, if there is a future docker-registry snap it's plugs and slots live in a namespace under `docker-registry:foo` so the docker snap can have `docker:docker-registry-certificates: and the docker-registry snap could have a completely separate `docker-registry:docker-registry-certificates` plug/slot and they don't conflict with each other

tianon:
Ah, thank you for the lesson :bow:

SGTM :+1:

tianon:
Thank you! :+1:

PR #57, opened 2022-01-03
Add optional "raw-usb" plug

### LINKED ISSUE #56: Please add raw-usb plug
I need to connect a Conbee II USB-Stick to my docker container.
Host is ubuntu core. 

When I run 
`sudo snap connect docker:raw-usb :raw-usb`
I get
`error: snap "docker" has no plug named "raw-usb"`

Please add a raw-usb plug so that people can use raw usb inside docker.

tianon:
Closes #56

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -49,6 +49,7 @@ plugs:
   home:
     read: all
   removable-media:
+  raw-usb:
   support:
     interface: docker-support
   privileged:


snap/snapcraft.yaml
```diff
@@ -49,6 +49,7 @@ plugs:
   home:
     read: all
   removable-media:
+  raw-usb:
```
anonymouse64, line 52: not sure if this works or not, since the plugs declared "globally" under the global `plugs` get applied to:

* all apps/hooks which do not declare _any_ plugs 
* any apps/hooks which specifically declare this plug

So as it is, I think this would only apply to any hooks, since the `dockerd` and `docker` apps do not declare this plug and declare a different set of plugs. I have a question for the reporter though, so hold off on merging this first

PR #79, opened 2022-10-20
Update to 20.10.20

tianon:


tianon:
Welp, I do not know what's causing `unable to prepare context: unable to 'git clone' to temporary context directory: error fetching: /usr/lib/git-core/git-remote-https: error while loading shared libraries: libcurl-gnutls.so.4: cannot open shared object file: No such file or directory` - I thought I had it twice now, and still no dice.

sbe-arg:
> Welp, I do not know what's causing `unable to prepare context: unable to 'git clone' to temporary context directory: error fetching: /usr/lib/git-core/git-remote-https: error while loading shared libraries: libcurl-gnutls.so.4: cannot open shared object file: No such file or directory` - I thought I had it twice now, and still no dice.

Have you tried different versions of compose instead of the latest one?

tianon:
Sorry, I'm not sure I understand?  Why would a different version of Compose give a different result for running `git` via `docker build`? :sweat_smile:

snap/snapcraft.yaml
```diff
@@ -170,6 +170,7 @@ parts:
     stage-packages:
       - aufs-tools
       - git
+      - libcurl3-gnutls
       - libltdl7
       - pigz
       - xz-utils
```
sbe-arg, line 176: Unsure wich one but I bet the problem is either 2 packages are conflicting with eachother at the symlink level or is missing libssl-dev

sbe-arg:
> Sorry, I'm not sure I understand? Why would a different version of Compose give a different result for running `git` via `docker build`? sweat_smile

I saw a few weeks ago similar stuff while compose trying a different versions helped.
I spend a bit more time understanding the snap build process here I think the problem is package related to the moby changes upstream and buildx.
Once docker is built lacks ssl (possibly) or gnu symlink and fails to run git-remote-https successfully.

sbe-arg:
@tianon if you accept the checks runs in this https://github.com/docker-snap/docker-snap/pull/80 pr I can help you debug it.

sbe-arg:
can you try adding `libcurl3-gnutls` here: https://github.com/docker-snap/docker-snap/blob/e8598c9438640f12eb9b629aaf209a0fd3f9ab25/snap/snapcraft.yaml#L271 ?

PR #88, opened 2023-02-28
compose-v2: Cherrypick "fix CVE-2022-27664 and CVE-2022-32149"

xnox:


snap/snapcraft.yaml
```diff
@@ -292,6 +292,8 @@ parts:
     # https://github.com/docker/compose/releases
     source-commit: v2.5.0
```
tianon, line 293: Wouldn't it be better to update this to [v2.16.0](https://github.com/docker/compose/releases/tag/v2.16.0) instead of cherry-picking individual Go dep changes?

xnox: i'm doing a minimal change update for now. And fighting packaging issues in the snapcraft.yaml as well.

xnox: And I don't have the confidence to bump to new major version, of such a large jump. Someone else is looking into it after me.

xnox:
i would like to merge this and put it in the candidate channel. And i will try to run the full spread test from there and some manual testing. and see if i can release it from there then.

xnox:
Released to stable.

PR #90, opened 2023-03-08
Add nvidia support

jocado:
This adds support for containers to access NVIDIA GPUs for acceleration

https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/overview.html

Adds related configuration and components from:
- nvidia-container-toolkit
- libnvidia-container

Also adds content interface plug for graphics-core22 provided by nvidia-core22 snap, containing required nvidia userspace libs.

Relates to: https://github.com/docker-snap/docker-snap/issues/91

### DIFF
--- a/snap/hooks/connect-plug-graphics-core22
+++ b/snap/hooks/connect-plug-graphics-core22
@@ -1,0 +1,15 @@
+#!/bin/bash -e
+
+# Ensure nvidia support setup correctly #
+if snapctl is-connected nvidia-core22:graphics-core22 ; then
+
+    # Generate the CDI config #
+    $SNAP/usr/bin/nvidia-ctk cdi generate --nvidia-ctk-path "$SNAP/usr/bin/nvidia-ctk" --output="$SNAP_DATA/etc/cdi/nvidia.yaml"
+
+    # Generate the dockerd runtime config #
+    $SNAP/usr/bin/nvidia-ctk runtime configure --runtime=docker --runtime-path "$SNAP/usr/bin/nvidia-ctk" --config "$SNAP_DATA/config/daemon.json"
+
+    # Restart dockerd to reflect any changes #
+    snap restart "$SNAP_NAME.dockerd"
+
+fi
--- a/snap/hooks/post-refresh
+++ b/snap/hooks/post-refresh
@@ -5,3 +5,19 @@ if [ ! -f "$SNAP_DATA/config/daemon.json" ]; then
     mkdir -p "$SNAP_DATA/config"
     cp "$SNAP/config/daemon.json" "$SNAP_DATA/config/daemon.json"
 fi
+
+# Ensure nvidia support setup correctly #
+if snapctl is-connected opengl ; then
+    if snapctl is-connected nvidia-core22:graphics-core22 ; then
+
+        # Generate the CDI config #
+        $SNAP/usr/bin/nvidia-ctk cdi generate --nvidia-ctk-path "$SNAP/usr/bin/nvidia-ctk" --output="$SNAP_DATA/etc/cdi/nvidia.yaml"
+
+        # Generate the dockerd runtime config #
+        $SNAP/usr/bin/nvidia-ctk runtime configure --runtime=docker --runtime-path "$SNAP/usr/bin/nvidia-ctk" --config "$SNAP_DATA/config/daemon.json"
+
+        # Restart dockerd to reflect any changes #
+        snap restart "$SNAP_NAME.dockerd"
+
+    fi
+fi
--- a/snap/hooks/post-refresh
+++ b/snap/hooks/post-refresh
@@ -7,17 +7,15 @@ if [ ! -f "$SNAP_DATA/config/daemon.json" ]; then
 fi
 
 # Ensure nvidia support setup correctly #
-if snapctl is-connected opengl ; then
-    if snapctl is-connected nvidia-core22:graphics-core22 ; then
+if snapctl is-connected nvidia-core22:graphics-core22 ; then
 
-        # Generate the CDI config #
-        $SNAP/usr/bin/nvidia-ctk cdi generate --nvidia-ctk-path "$SNAP/usr/bin/nvidia-ctk" --output="$SNAP_DATA/etc/cdi/nvidia.yaml"
+    # Generate the CDI config #
+    $SNAP/usr/bin/nvidia-ctk cdi generate --nvidia-ctk-path "$SNAP/usr/bin/nvidia-ctk" --output="$SNAP_DATA/etc/cdi/nvidia.yaml"
 
-        # Generate the dockerd runtime config #
-        $SNAP/usr/bin/nvidia-ctk runtime configure --runtime=docker --runtime-path "$SNAP/usr/bin/nvidia-ctk" --config "$SNAP_DATA/config/daemon.json"
+    # Generate the dockerd runtime config #
+    $SNAP/usr/bin/nvidia-ctk runtime configure --runtime=docker --runtime-path "$SNAP/usr/bin/nvidia-ctk" --config "$SNAP_DATA/config/daemon.json"
 
-        # Restart dockerd to reflect any changes #
-        snap restart "$SNAP_NAME.dockerd"
+    # Restart dockerd to reflect any changes #
+    snap restart "$SNAP_NAME.dockerd"
 
-    fi
 fi
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -35,12 +35,6 @@ base: core22
 confinement: strict
 assumes: [snapd2.50]
 
-environment:
-  # For nvidia support #
-  LD_LIBRARY_PATH:    $SNAP/graphics/lib:${SNAP}/lib/:${SNAP}/lib/${SNAPCRAFT_ARCH_TRIPLET}:${SNAP}/usr/lib/:${SNAP}/usr/lib/${SNAPCRAFT_ARCH_TRIPLET}
-  LIBGL_DRIVERS_PATH: $SNAP/graphics/dri
-  LIBVA_DRIVERS_PATH: $SNAP/graphics/dri
-
 layout:
   /etc/docker:
     bind: $SNAP_DATA/etc/docker
@@ -55,6 +49,10 @@ layout:
 environment:
   GIT_EXEC_PATH: "$SNAP/usr/lib/git-core"
   GIT_TEMPLATE_DIR: "$SNAP/usr/share/git-core/templates"
+  # For nvidia support #
+  LD_LIBRARY_PATH:    $SNAP/graphics/lib:${SNAP}/lib/:${SNAP}/lib/${SNAPCRAFT_ARCH_TRIPLET}:${SNAP}/usr/lib/:${SNAP}/usr/lib/${SNAPCRAFT_ARCH_TRIPLET}
+  LIBGL_DRIVERS_PATH: $SNAP/graphics/dri
+  LIBVA_DRIVERS_PATH: $SNAP/graphics/dri
 
 plugs:
   home:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -222,7 +222,17 @@ parts:
     source: https://github.com/NVIDIA/nvidia-container-toolkit.git
     source-tag: v1.12.0
     source-depth: 1
+    override-pull: |
+      [ "${CRAFT_TARGET_ARCH}" != "amd64" ] && \
+        [ "${CRAFT_TARGET_ARCH}" != "arm64" ] && \
+        exit 0
+
+      craftctl default
     override-build: |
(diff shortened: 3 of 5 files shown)


xnox:
Is nvidia-core22 content provider really needed? I think in some other snap we already pre-built support for this from source, without using the full graphics-core22 interface. Possibly in lxd or microk8s, will check.

Also security wise, lxd microk8s are much better choices. Are there any specific reasons to use very old docker snap, instead of better maintained lxd or microk8s? this docker snap really needs to get upgraded to never base and newer version.

snap/hooks/connect-plug-graphics-core22
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
+
+# Ensure nvidia support setup correctly #
+if snapctl is-connected nvidia-core22:graphics-core22 ; then
```
xnox, line 4: it's not ok to mix bases like this. nvidia-core22 is only meant to be used strictly with snaps that are also base:core22.

jocado: Hmm, ok.  I'm not sure how to work around that then.  We are using the docker snap on both UC20 and UC22.

jocado:
> Is nvidia-core22 content provider really needed? I think in some other snap we already pre-built support for this from source, without using the full graphics-core22 interface. Possibly in lxd or microk8s, will check.

It seems to be required.  Without it, I get the following when trying to run a container with gpu enabled:
```
nvidia-container-cli: initialization error: load library failed: libnvidia-ml.so.1: cannot open shared object file: no such file or directory: unknown.
```
> Also security wise, lxd microk8s are much better choices. Are there any specific reasons to use very old docker snap, instead of better maintained lxd or microk8s? this docker snap really needs to get upgraded to never base and newer version.

We are bound to docker in the cases we are working on, currently.  This is for a variety of "internal" reasons, and so is a hard requirement for us right now.

snap/hooks/connect-plug-graphics-core22
jocado: Would it be true to say then, for a docker snap with nvidia support it would have to be built for each required UC version base ?

I guess that could potentially still be solved by git branch mapped to snap channel.

xnox:
I would feel better to start docker track on core22 base, and maybe making it the default. Because we can end up with runtime incompatibilities between nvidia-ml library and other shared libraries from the base.

snap/hooks/connect-plug-graphics-core22
xnox: no, not quite. It probably is best to compile the require tool, as a part inside the docker snap, and vendor the require library it needs. Such that it is self contained, and can upgrade the tool and libraries to be compatible with what it has. And that way it should be self contained.

snap/snapcraft.yaml
```diff
@@ -207,6 +222,45 @@ parts:
...
+
+      make ${PLATFORM_DIST}
+
+      dpkg-deb -x ${SNAPCRAFT_PART_BUILD}/dist/${PLATFORM_DIST}/${SNAPCRAFT_TARGET_ARCH}/libnvidia-container1_*.deb ${SNAPCRAFT_PART_INSTALL}
+      dpkg-deb -x ${SNAPCRAFT_PART_BUILD}/dist/${PLATFORM_DIST}/${SNAPCRAFT_TARGET_ARCH}/libnvidia-container-tools_*.deb ${SNAPCRAFT_PART_INSTALL}
+    build-packages:
+      - make
+    # Recursive - could also use docker.io build-packages #
+    build-snaps:
+      - docker
```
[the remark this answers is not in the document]
xnox: Also interesting that lxd does not use nvidia-ctk bur only uses nvidia-container-cli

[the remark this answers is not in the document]
jocado: Sure - I will take a look at that, and see what's possible :+1:

snap/hooks/connect-plug-graphics-core22
xnox: also the nvidia-core22 libraries may not be compatible with the contents of the containers, I do expect containers to potentially contain their own userspace libraries. And I expected for host side to not need any of the sort to get things going. Will study the nvidia-ctk more to see what can be done there.

snap/snapcraft.yaml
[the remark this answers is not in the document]
jocado: > Also interesting that lxd does not use nvidia-ctk bur only uses nvidia-container-cli

So far, everything I've included is pretty much just according to the various documentation I've found on the NIVIDIA and Docker web sites.  It's possible it could be paired down a bit and still work, but following documented installation seems like a good starting point at least.

snap/hooks/connect-plug-graphics-core22
jocado: @xnox I had a quick go at switching to a core22 base.  Not too many changes required, and it does seem to work.  I haven't included the changes in this PR yet, but for ref they are in a separate branch, called `core22-nvidia-support`

Here's the diff:
- https://github.com/jocado/docker-snap/compare/nvidia-support...jocado:docker-snap:core22-nvidia-support

xnox (changes requested):
At the moment graphics-core22 interface is not yet declared as stable.
Thus I would want to wait a little bit before starting to use it.
Especially since a command-chain wrapper will be provided.

see https://github.com/xnox/nvidia-core22/pull/2/files

xnox:
Using docker snap, during docker snap build is not acceptable. All other parts, set explicit options to by-pass and ignore lack of a docker container.

Similarly new nvidia parts may not use a nested docker container for the build; given the snap build is done in a constraint environment as is.

jocado:
ok.  I was starting to dig into why `build-snaps: [docker]` was not working.

ok, so the nvidia parts will need to re-implement the build without then.  That'a a bit of a shame, as it means more code to maintain, but I will look into ti.

jocado:
@xnox  Just a very quick look so far, but it looks like build solution in those nvidia repos only supported docker based builds.

I see two options, unless you have any other suggestions:
- Re-implement the docker based build using using logic in the part and accept the ongoing maintenance of it
  - I don't know how complex or not it would be yet
- use the already build packages from the nvidia apt repo
  - Probably much less work and ongoig maintenance, but providence supply chain are less well understood
 
Thoughts ?

jocado:
FYI - I have reworked the `libnvidia-container` inspirded by the LXD snap, as originally suggested, it was pretty straight forward.

Just working on the nvidia-container-toolkit part now.

jocado:
The nvidia parts should be looking a bit better now.

jocado:
FYI - I haven't included the wrapper for nvidia-container-cli from the lxd snap at this point.  It _seems_ like it could be useful perhaps for the classic use case, but wanted to discuss with you @xnox to see what you thought about it first.

snap/hooks/post-refresh
```diff
@@ -5,3 +5,19 @@ if [ ! -f "$SNAP_DATA/config/daemon.json" ]; then
...
+    if snapctl is-connected nvidia-core22:graphics-core22 ; then
+
+        # Generate the CDI config #
+        $SNAP/usr/bin/nvidia-ctk cdi generate --nvidia-ctk-path "$SNAP/usr/bin/nvidia-ctk" --output="$SNAP_DATA/etc/cdi/nvidia.yaml"
+
+        # Generate the dockerd runtime config #
+        $SNAP/usr/bin/nvidia-ctk runtime configure --runtime=docker --runtime-path "$SNAP/usr/bin/nvidia-ctk" --config "$SNAP_DATA/config/daemon.json"
+
+        # Restart dockerd to reflect any changes #
+        snap restart "$SNAP_NAME.dockerd"
```
xnox, line 20: what are the conditions for regenerating this data, and what does it depend on? Should it be rebuild each time there is change to nvidia.ko kernel driver? userspace libraries? Each time dockerd starts?

On every boot, nvidia-assemble assembles a potentially new driver, and then modprobes a new config. Should we generate this upon each dockerd daemon start up? to ensure it is correct, and restarting dockerd gets us the right config?

I'm not at all familiar with this, but if it is quick and ephemeral, I'd rather do this in like runtime-dir and do it on every dockerd startup. If at all possible.

snap/snapcraft.yaml
```diff
@@ -35,9 +35,18 @@ base: core22
 confinement: strict
 assumes: [snapd2.50]
 
+environment:
+  # For nvidia support #
+  LD_LIBRARY_PATH:    $SNAP/graphics/lib:${SNAP}/lib/:${SNAP}/lib/${SNAPCRAFT_ARCH_TRIPLET}:${SNAP}/usr/lib/:${SNAP}/usr/lib/${SNAPCRAFT_ARCH_TRIPLET}
+  LIBGL_DRIVERS_PATH: $SNAP/graphics/dri
+  LIBVA_DRIVERS_PATH: $SNAP/graphics/dri
+
```
xnox, line 43: @xnox need to check if this is still valid with nvidia-core22 in edge.

snap/snapcraft.yaml
```diff
@@ -58,6 +67,11 @@ plugs:
     privileged-containers: true
   docker-cli:
     interface: docker
+  # For nvidia userspace libs support #
+  graphics-core22:
+    interface: content
+    target: $SNAP/graphics
+    default-provider: nvidia-core22
```
xnox, line 74: @xnox check if this is still correct target

snap/snapcraft.yaml
```diff
@@ -200,6 +217,47 @@ parts:
...
+  nvidia-container-toolkit:
+    plugin: go
+    source: https://github.com/NVIDIA/nvidia-container-toolkit.git
+    source-tag: v1.12.0
+    source-depth: 1
+    override-build: |
+      craftctl default
+
+      # From nvidia-container-toolkit package postinst
+      (cd "${CRAFT_PART_INSTALL}/bin" && ln -s nvidia-container-runtime-hook nvidia-container-toolkit)
```
xnox, line 229: is this compat / legacy name? or does the binary behave different when called under different name?

xnox:
This looks very good now from building point of view. Need to double check if we should call things with a command-chain wrappers, or not. Cause maybe we should be surfacing these new binaries as snap commands. Not sure.

The points raised above are minor, as this is probably in a mergeable state already.

xnox:
```console
snapcraft remote-build --build-on amd64,arm64,armhf,s390x,ppc64el
```

fail for multiple reasons. It seems we cannot specify default-provider: nvidia-core22, as it doesn't exist on arches beyond amd64/arm64, and yet we cannot make that slot arch specific.

similarly, since there are no nvidia drivers on amd64, arm64, we should skip building nvidia parts, and hooks on those architectures.

sigh.

snap/snapcraft.yaml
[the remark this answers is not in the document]
jocado: I think this thread was resolved by my changes yesterday.

snap/snapcraft.yaml
jocado: To be honest, I don't know the answer, as per my comment int eh part it's from the upstream package postinst.

I will try and find out.  If I find nothing useful about it, we can always just take it out and see what breaks.

jocado:
> fail for multiple reasons. It seems we cannot specify default-provider: nvidia-core22, as it doesn't exist on arches beyond amd64/arm64, and yet we cannot make that slot arch specific.

ok, I will remove the default provider. I believe auto-connection should still work if the content identifier matches, and the desired snap that provides the interfaces is installed first.  Possibly the publisher of the nvidia-core22 snap would have to be change to Canonical [ so publisher matches for both consumer and producer ], not completely sure.  Did you plan to migrate the `nvidia-core*` snaps to Canonical publisher eventually ?

> similarly, since there are no nvidia drivers on amd64, arm64, we should skip building nvidia parts, and hooks on those architectures.

Makes sense, I will take a look at that.

snap/snapcraft.yaml
jocado: Turns out this was renamed at some point, and seems to be for compatibility.

Ref: https://github.com/NVIDIA/nvidia-container-toolkit/commit/8a56671d18a3f1953ea973b1b744bc9cb97b72b9

Therefor, as it;s internal to the snap at this point, it should be find to remove, which I've done.

jocado:
I have made the nvidia parts amd64 and arm64 target only.

I will try and look at any remaining point tomorrow.

xnox:
Imho this is enough for edge channel, will need to be retested with new nvidia-core22 (it restructured things) and I will need to retest myself on my instances. And For me it is easiest to do that form the snapstore.

And I will probably not land this into stable until after 2.59 snapd lands in stable anyway.

xnox (approved):

jocado:
Great - thanks @xnox 

I was intending to do further functional testing today, with latest version of `nvidia-assemble` and `nvidia-core22`.

I will discuss any further points in new issues or PR etc :+1:

jocado:
> Possibly the bin/ wrapper should be used when calling Nvidia binaries without all the Environment overrides.
That's exactly what I was thinking :+1:

PR #92, opened 2023-03-10
Use source-tag, instead of commit, to avoid rebuilds

xnox:


xnox:
Will also require https://github.com/canonical/snapcraft-poller-script/pull/51 to land

PR #93, opened 2023-03-13
Switch to core22 base

jocado:
Make required changes to enable snap build with core22 base

Relates to: https://github.com/docker-snap/docker-snap/issues/65

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -186,7 +186,7 @@ parts:
 
       install -d "$SNAPCRAFT_PART_INSTALL/bin"
       install -t "$SNAPCRAFT_PART_INSTALL/bin" bin/containerd* bin/ctr
-    go-channel: &gochannel '1.18/stable'
+    build-snaps: *go
     build-packages:
       - make
       - libbtrfs-dev
@@ -201,7 +201,7 @@ parts:
 
       install -d "$SNAPCRAFT_PART_INSTALL/bin"
       install -T runc "$SNAPCRAFT_PART_INSTALL/bin/runc"
-    go-channel: *gochannel
+    build-snaps: *go
     build-packages:
       - libapparmor-dev
       - libseccomp-dev
@@ -285,7 +285,7 @@ parts:
 
       install -d "$SNAPCRAFT_PART_INSTALL/usr/libexec/docker/cli-plugins"
       install -T buildx "$SNAPCRAFT_PART_INSTALL/usr/libexec/docker/cli-plugins/docker-buildx"
-    go-channel: *gochannel
+    build-snaps: *go
 
   compose-v2:
     plugin: go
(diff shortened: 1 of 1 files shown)


xnox (approved):
But let's see how our ci will react to things.

jocado:
I've pushed a change to try and fix the previous workflow failure, but it seems you have to approve each run @xnox

jocado:
I guess its because the branch is in my own fork

jocado:
Snap was built at least.

I'm not 100% how to interpret the failure: https://github.com/docker-snap/docker-snap/actions/runs/4409275831/jobs/7725502411#step:10:6

Do you have any thoughts @xnox ?

Worth noting that the snap I built locally ran fine on a UC22 host, tested with hello-world docker image.

I suppose if this was my own CI permission in, I would add some debug to the workflow and run it again.

tianon:
A core22 snap is not supported on an Ubuntu 20.04 system, right?  If it is, then this failure is worth digging more into, but if it isn't then you probably just want to update this line to `ubuntu-22.04` instead:

https://github.com/docker-snap/docker-snap/blob/975f02462d78b74b96da4629dff480d14ec57542/.github/workflows/smoke-test.yml#L14

xnox:
> A core22 snap is not supported on an Ubuntu 20.04 system, right?  If it is, then this failure is worth digging more into, but if it isn't then you probably just want to update this line to `ubuntu-22.04` instead:
> 
> https://github.com/docker-snap/docker-snap/blob/975f02462d78b74b96da4629dff480d14ec57542/.github/workflows/smoke-test.yml#L14

We could update to 22.04 host base for workflows. But in theory it should work. We have Lxd snap built on any base, and running on any host. Including when Lxd snap was base:core20 and when it moved to base:core22.

tianon:
Ah, so the failure probably _is_ worth investigating then, because that's even the most basic use case of `docker run hello-world` failing :sweat_smile:

xnox:
> Ah, so the failure probably _is_ worth investigating then, because that's even the most basic use case of `docker run hello-world` failing sweat_smile

i think it might be fun things between focal host, core22 confinement, and changes in apparmor_parser --version which tries to parse things, and gets a denial to access /abi/ subdir, that is there, but not allowed to by the docker snapd interface.

This probably works on UC because we don't ship /abi/ directory on UC and nothing tries to access it either. Fun things.

Somewhat wonder if we should try upgrade to core20 first, cause fixing this issue may require upgrades to docker itself, and/or changes to the snapd interfaces.

xnox (changes requested):
Block merging until we get feedback from interface maintainers and apparmor maintainers.

jocado:
@xnox Just to join the dots - what was the feedback from interface maintainers and apparmor maintainers ?

PR #98, opened 2023-03-14
Upgrade 20.10.20 take3

xnox:


xnox:
@tianon @sbe-arg thank you for trying this before.

As far as I can tell, the snap is being missbuilt.
When git-ubuntu snap stages git package into the snap it gains in readelf:
```
[Requesting program interpreter: /snap/core20/current/lib64/ld-linux-x86-64.so.2]
Dynamic section at offset 0x1c2000 contains 32 entries:
  Tag        Type                         Name/Value
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/../x86_64-linux-gnu:/snap/core20/current/lib/x86_64-linux-gnu
```
on its binaries, i.e. the git-remote-http binary.

But the docker snap instead doesn't have that. As if patchelf has not been executed. Or stripped by something.

Thus the binary doesn't use libraries from the base, nor from docker snap itself, and expects the host to provide those libraries, which our workflow runner doesn't have.

So something is changed in the build that strips rpath and interpret setting from all the staged/installed binaries.

xnox:
getting somewhere, something somewhere eats environment variables.

dockerd-patches/snappy-buildkit-git-environ.patch
```diff
@@ -1,22 +1,14 @@
...
 --- a/vendor/github.com/moby/buildkit/source/git/gitsource.go
 +++ b/vendor/github.com/moby/buildkit/source/git/gitsource.go
-@@ -598,12 +598,11 @@ func git(ctx context.Context, dir, sshAuthSock, knownHosts string, args ...strin
- 		cmd.Stdin = nil
- 		cmd.Stdout = io.MultiWriter(stdout, buf)
+@@ -613,6 +613,9 @@ func git(ctx context.Context, dir, sshAuthSock, knownHosts string, args ...strin
  		cmd.Stderr = io.MultiWriter(stderr, errbuf)
--		cmd.Env = []string{
--			"PATH=" + os.Getenv("PATH"),
-+		cmd.Env = append(append([]string{}, os.Environ()...), []string{
```
tianon, line 14: Did you leave all of `os.Environ()` in here on purpose?

(there's probably no need to do the later `os.Getenv` calls if so, but I think this list was intentionally limited to mitigate different environment variables that `git` might pull in that would affect the way buildkit invokes it :grimacing:)

xnox: @tianon it's a diff of a diff
```console
- +		cmd.Env = append(append([]string{}, os.Environ()...), []string{
```
meaning the line of append all of os.Environ, is removed from this patch.

see https://github.com/xnox/docker-snap/blob/7a5e055ffdf5c48efc5ea3a1884c79e98ceaf17b/dockerd-patches/snappy-buildkit-git-environ.patch how the patch actually looks like now.

tianon: Doh, yes, sorry; it has been a long week. :facepalm:

Thank you for explaining :bow:

PR #105, opened 2023-03-17
Fix the graphics paths used by the nvidia support

jocado:
Update the environment paths used for the graphics content interface, following changes in the provider


jocado:
@xnox I will look at using the wrapper, but in the meantime this just get the nvidia support working again with the current graphics interface

snap/snapcraft.yaml
```diff
@@ -50,9 +50,9 @@ environment:
   GIT_EXEC_PATH: "$SNAP/usr/lib/git-core"
   GIT_TEMPLATE_DIR: "$SNAP/usr/share/git-core/templates"
   # For nvidia support #
-  LD_LIBRARY_PATH:    $SNAP/graphics/lib:${SNAP}/lib/:${SNAP}/lib/${SNAPCRAFT_ARCH_TRIPLET}:${SNAP}/usr/lib/:${SNAP}/usr/lib/${SNAPCRAFT_ARCH_TRIPLET}
-  LIBGL_DRIVERS_PATH: $SNAP/graphics/dri
-  LIBVA_DRIVERS_PATH: $SNAP/graphics/dri
+  LD_LIBRARY_PATH:    $SNAP/graphics/lib/${SNAPCRAFT_ARCH_TRIPLET}:${SNAP}/lib/:${SNAP}/lib/${SNAPCRAFT_ARCH_TRIPLET}:${SNAP}/usr/lib/:${SNAP}/usr/lib/${SNAPCRAFT_ARCH_TRIPLET}
+  LIBGL_DRIVERS_PATH: $SNAP/graphics/lib/${SNAPCRAFT_ARCH_TRIPLET}/dri
+  LIBVA_DRIVERS_PATH: $SNAP/graphics/lib/${SNAPCRAFT_ARCH_TRIPLET}/dri
```
xnox, line 55: shouldn't these be `/graphics/usr/lib/`throughout? given the nvidia-core22 snap only has things in `/usr/lib` and i don't want to use incorrect paths.

jocado: It seems not.  As noted [here](https://github.com/xnox/graphics-core22-samples/issues/4#issuecomment-1475183502) , the resulting bind mounts are at the top level of the target.  So `$SNAP/usr/lib` becomes `graphics/lib`.

So, this does seem correct, and I have tested it locally too to make sure.

xnox (approved):

PR #111, opened 2023-03-21
Remove docker compose v1

zhsj:
It's officially EOL.

See https://github.com/docker/compose/commit/30fcb72c

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -111,11 +111,12 @@ apps:
       - docker-daemon
 
   compose:
-    command: bin/docker-compose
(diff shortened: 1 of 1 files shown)


tianon (approved):
vigorously LGTM :smile:

xnox:
I think there are cases where new one doesn't work for people, until we bump golang to a newer one. (something something proxy/networking whatever, everything is broken, must revert to python one).

So I would be happy to do this, as long as we are on latest v2 compose, which is built with the latest golang.

xnox:
Given https://github.com/docker/roadmap/issues/257#issuecomment-1427024325 and https://github.com/docker-snap/docker-snap/pull/118 we can merge this.

Issue #81, opened 2022-11-01
python installation issues on docker snap


greatquux:
Install docker snap and create an Ubuntu 22.10 container, and try to install python, you will get an error:
Setting up python3.10-minimal (3.10.7-1) ...
[Errno 13] Permission denied: '/usr/lib/python3.10/__pycache__/__future__.cpython-310.pyc.139786761293248'dpkg: error processing package python3.10-minimal (--configure):
 installed python3.10-minimal package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 python3.10-minimal
E: Sub-process /usr/bin/dpkg returned an error code (1)

See
https://forums.docker.com/t/bug-on-apt-install-permission-denied/100196/6
for discussion.  Apparently only a problem with the snap version.

greatquux:
BTW once I did find where my daemon.json file was, I switched to using the vfs engine and did not have any further problems with python.  But I think the issue should be fixed for overlay2.

mgajda:
The problem still persists, and [Docker recommends installing from `.deb` package repository as a workaround](https://docs.docker.com/engine/install/ubuntu/#set-up-the-repository).

plars:
For now, you can work around this by using the vfs storage-driver instead:
- edit /var/snap/docker/current/config/daemon.json and replace "overlay2" with "vfs" for the storage-driver
- run `sudo snap restart docker`

kurapov-peter:
Observing the same issue here

zhsj:
The issue can be easily trigger with following command

```
docker run --rm ubuntu:devel sh -c 'mkdir -p /usr/lib/python3/__pycache__; echo 1 > /usr/lib/python3/__pycache__/x.pyc.1'
```

This seems to caused by the apparmor policy.

Maybe this one?

https://github.com/snapcore/snapd/blob/19c123f2b3f67c6e79e1f0a2af7c0721194b3a70/interfaces/apparmor/template.go#L108
```
deny /usr/lib/python3*/{,**/}__pycache__/**.pyc.[0-9]* w,
```

But I'm out of idea why the kernel doesn't use the path which is translated after overlayfs.

zhsj:
```
[...]

PR #113, opened 2023-03-22
Add regression test for installing python3

zhsj:
Updates #81


xnox:
If/when snapd/apparmor is fixed, i think we will want to add this test case.

xnox:
This should be fixed by snapd now, and somehow need to retrigger the workflow.

zhsj:
> This should be fixed by snapd now, and somehow need to retrigger the workflow.

I probably should close this PR, as we can't add such tests with Github Action. The error is only triggered by specific kernel version, while we can't change kernel version on GitHub Action.

PR #117, opened 2023-03-22
Bump GitHub Action image to ubuntu 22.04

zhsj:
This needs snapcore/action-build@v1.1.2 to fix network issue. See https://github.com/snapcore/action-build/issues/42

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -11,13 +11,27 @@ defaults:
 jobs:
   smoke-test:
     name: Snapcraft
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v2
 
+      - name: Build
+        # https://github.com/snapcore/action-build/releases
+        uses: snapcore/action-build@v1
+        with:
+          snapcraft-args: snap --output docker.snap
+
+      - name: Upload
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -11,13 +11,27 @@ defaults:
...
   smoke-test:
     name: Snapcraft
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v2
 
+      - name: Build
+        # https://github.com/snapcore/action-build/releases
+        uses: snapcore/action-build@v1
```
xnox, line 20: Why is this action downgraded from previous v1.1.1 to v1? don't you want to upgrade to v1.1.2 here?

zhsj: We don't need to pin it to a micro release. https://github.com/snapcore/action-build#readme suggests using `v1` as well.

xnox: ah, ok.

However the build still fails. I wonder if we are missing something to allow unified cgroups to work, or if nested unified cgroups prevent running docker nested.

zhsj: Yeah, I'm moving it to draft. I'll work on it this week.

PR #119, opened 2023-03-28
Drop default storage-driver, it is the same as autodetected/picked on…

xnox:
…e anyway

Fixes: https://github.com/docker-snap/docker-snap/issues/112


config/daemon.json
```diff
@@ -1,4 +1,3 @@
 {
     "log-level":        "error",
```
tianon, line 2:
```suggestion
    "log-level":        "error"
```

PR #125, opened 2023-04-25
NVIDIA support fixes

jocado:
Somehow the interface connection test in the hooks was not configured correctly, and so the hooks were not running.  This fixes that.

### DIFF
--- a/snap/hooks/connect-plug-graphics-core22
+++ b/snap/hooks/connect-plug-graphics-core22
@@ -1,7 +1,7 @@
 #!/bin/bash -e
 
 # Ensure nvidia support setup correctly #
-if snapctl is-connected nvidia-core22:graphics-core22 ; then
+if snapctl is-connected graphics-core22 ; then
 
     # Generate the CDI config #
     $SNAP/usr/bin/nvidia-ctk cdi generate --nvidia-ctk-path "$SNAP/usr/bin/nvidia-ctk" --output="$SNAP_DATA/etc/cdi/nvidia.yaml"
--- a/snap/hooks/install
+++ b/snap/hooks/install
@@ -13,7 +13,7 @@ mkdir -p "$SNAP_DATA/etc/docker"
 mkdir -p "$SNAP_DATA/etc/cdi"
 
 # Ensure nvidia support setup correctly #
-if snapctl is-connected nvidia-core22:graphics-core22 ; then
+if snapctl is-connected graphics-core22 ; then
 
     # Generate the CDI config #
     $SNAP/usr/bin/nvidia-ctk cdi generate --nvidia-ctk-path "$SNAP/usr/bin/nvidia-ctk" --output="$SNAP_DATA/etc/cdi/nvidia.yaml"
--- a/snap/hooks/post-refresh
+++ b/snap/hooks/post-refresh
@@ -7,7 +7,7 @@ if [ ! -f "$SNAP_DATA/config/daemon.json" ]; then
 fi
 
 # Ensure nvidia support setup correctly #
-if snapctl is-connected nvidia-core22:graphics-core22 ; then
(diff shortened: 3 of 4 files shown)


jocado:
@xnox I've had a go at improving the NVIDIA support - PTAL

bin/nvidia-container-toolkit
```diff
@@ -2,12 +2,38 @@
...
+    return 1
+
+}
+
+# Ensure nvidia support setup correctly, and only if hardware is preset and correct #
 if snapctl is-connected graphics-core22 ; then
 
-    # Only set this up if the system has an nvidia card available #
+    # If module is loaded - hardware _should_ be present #
     grep -Eq "^nvidia" /proc/modules || exit 0
```
xnox, line 33: this is not needed and racy, as module may not be loaded yet, whilst the device_wait function would wait for that to happen.

bin/nvidia-container-toolkit
```diff
@@ -2,12 +2,38 @@
 
 set -eu
 
-# Ensure nvidia support setup correctly #
+device_wait() {
```
xnox, line 5: A bit sad we need to do all of this, because we cannot simply be `After=snap.nvidia-assemble.service` (or some such)

bin/nvidia-container-toolkit
jocado: The main reason I left this in was because it seemed like a nice way of simply skipping most of the logic if there was no chance of the hardware being present at all.  But, perhaps as you point out, out that assumption could be flawed.

Without this, the `dockerd` service would always wait for the `nvidia-container-toolkit` to finish [ 30 secs with current wait log timing settings ] on systems that simply don't have nvidia hardware.

Given the intent here is to not artificially delay setup on a system with no hope of an nvidia device appearing, can you think of any alternative/better way that would not require any overly privileged snap interface ?

xnox: 1. device needs to be present (`lspci -d 10de:` should be non-empty)
2. On classic - kernel module needs to be loaded & /hostfs libraries should be available and usable
3. On core - nvidia-assemble needs to be installed and running (i think some other snaps export a socket over an interface, or just export an empty interface, to allow others to connect and thus observe that nvidia-assemble is installed/running).

However, I hope that graphics-core22 is not automatically installed and connected, thus maybe installing and connecting that should be enough as a wait?

Can ponder and improve that later.

README.md
```diff
@@ -47,6 +47,18 @@ Docker should function normally, with the following caveats:
 
 * Additional certificates used by the Docker daemon to authenticate with registries need to be located in `/var/snap/docker/common/etc/certs.d` instead of `/etc/docker/certs.d`.
 
+## NVIDIA support
```
xnox, line 50: We haven't yet make it work on Classic in expected ways (using hostfs libraries), thus maybe we should ommit this from README.md for now, or like mention that it's targeting Ubuntu Core 22 only for now.

xnox: Or maybe nobody reads this anyway =)

xnox (approved):
I think we can merge this as is, and iterate to improve.

bin/nvidia-container-toolkit
jocado: @xnox 

I have been thinking abou tthis, and was just about to commit another change to make the NVIDIA support "opt in" only, via snap config.  I think that address most concerns for now.

Shall I add it and see what you think ?

xnox: not sure. it still feels like an opt-out to me. If you have hardware, and drivers, and connect them up, one likely wants to use them.

jocado: ok, cool, that work for me - I will propose an opt-out approach in a separate PR :+1:

PR #128, opened 2023-04-28
NVIDIA opt-out

jocado:
Add the ability to completely op-out of setup up NVIDIA support, whether the hardware is present or not, by setting snap config.

By setting `nvidia-support.disabled=true` the daemon config file will have the nvidia runtime config remove, the CDI config for nvidia will be removed too, and the `nvidia-container-toolkit` service will be skipped.

This also improves the nvidia hardware detection, by checking for the presence of a PCI device with the relevant ID.

Fixes a bug effecting first install of this snap on a system.

### DIFF
--- a/bin/nvidia-container-toolkit
+++ b/bin/nvidia-container-toolkit
@@ -33,6 +33,12 @@ NVIDIA_SUPPORT_DISABLED="$(snapctl get nvidia-support.disabled)"
 # Ensure nvidia support setup correctly, and only if hardware is preset and correct #
 if snapctl is-connected graphics-core22 ; then
 
+    # Connection hooks are run early - copy the config file from $SNAP into $SNAP_DATA if it doesn't exist
+    if [ ! -f "$SNAP_DATA/config/daemon.json" ]; then
+        mkdir -p "$SNAP_DATA/config"
+        cp "$SNAP/config/daemon.json" "$SNAP_DATA/config/daemon.json"
+    fi
+
     # Check if hardware is present - just exit if not #
     lspci -d 10de: | grep -q 'NVIDIA Corporation' || exit 0
     echo "NVIDIA hardware detected: $(lspci -d 10de:)"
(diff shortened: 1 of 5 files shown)


jocado:
@xnox While I had some time I too a look at adding opt-out via snap config.  IMO this make this safer for eventual stable release.

I also took the chance to improve the NV hardware detection [ in the way you suggested ], and luckily found and fixed a bug with first install.

PTAL when you get a chance.

bin/nvidia-container-toolkit
```diff
@@ -26,15 +26,26 @@ device_wait() {
 
 }
 
+# Just exit if NVIDIA support is disabled #
+NVIDIA_SUPPORT_DISABLED="$(snapctl get nvidia-support.disabled)"
```
xnox, line 30: Thinking more about this, this adds a second toggle to turn Nvidia support off. Because, previous pull request already added that (potentially without cleanup) indirectly.

Since nvidia-container-toolkit is now a service, one can do `snap services disable docker.nvidia-container-toolkit` which feels like a natural thing to do, if that is the service one doesn't want from the docker snap.

I don't know if the snap is notified about such change and if we should key on if `snapctl services` is enabled or not.

The changes themselves are good. So could be merged, but I do wonder if we can get away with just service enabled/disabled toggle (default enabled).

xnox (approved):
The cleanup is good, restarts is good, hw detection is good. I am backtracking on the opt out. And pondering if our one shot service should remainafterexit to be able to restart it.

xnox:
It seems like disabling service is not notified to the snap via configure hook trigger. Sadness.

bin/nvidia-container-toolkit
jocado: Yes, I think the problem is, as you already notice, disable does give any usable signal which could trigger a cleanup.

jocado:
@xnox Thanks for the review :+1:

Unless you have any better suggestion right now, I vote to merge this for now as is, as it fixes a bug too.

I'd be happy to re-visit the "disable mechanism" topic in the future if one of us or anyone else has an idea to improve it.

jocado:
> @xnox Thanks for the review +1
> 
> Unless you have any better suggestion right now, I vote to merge this for now as is, as it fixes a bug too.
> 
> I'd be happy to re-visit the "disable mechanism" topic in the future if one of us or anyone else has an idea to improve it.

Hi @xnox  - apologies, I'm sure you're busy, but just wondering if saw my message above ?

Are you happy to merge as-is for now, especially as it fixes a bug, or would you prefer some changes first ?

Thanks.

PR #131, opened 2023-05-15
snap: arch-restrict all steps

xnox:
This should fix FTBFS on non-arm64|amd64.

### DIFF
--- a/patches/libnvidia-container/nvidia-0001-Fix-for-22.04-build.patch
+++ b/patches/libnvidia-container/nvidia-0001-Fix-for-22.04-build.patch
@@ -26,7 +26,7 @@ index 00d561e..8710bfd 100644
  LIB_LDLIBS_STATIC  += -l:libtirpc.a
  LIB_LDLIBS_SHARED  += -lpthread
 +else
-+LIB_CPPFLAGS       += $(shell pkg-config --cflags libtirpc)
++LIB_CPPFLAGS       += -DWITH_TIRPC $(shell pkg-config --cflags libtirpc)
 +LIB_LDLIBS_SHARED  += $(shell pkg-config --libs libtirpc)
  endif
  ifeq ($(WITH_SECCOMP), yes)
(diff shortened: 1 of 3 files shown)


xnox:
@jocado somehow the github actions builds are all failing now, and i'm not sure how or why.

They used to work fine.

Also i saw that nvidia stuff got downgraded from 13 releases back to 12, was that intentional or accidental? I cannot tell from the git log.

jocado:
> Also i saw that nvidia stuff got downgraded from 13 releases back to 12, was that intentional or accidental? I cannot tell from the git log.

Apologies - the revert was intentional, but I accidentally added it to another commit it seems.  The build seemed to be failing with that version, and I didn't have to to investigate the cause, so I postponed that version bump until later.

xnox:
> > Also i saw that nvidia stuff got downgraded from 13 releases back to 12, was that intentional or accidental? I cannot tell from the git log.
> 
> Apologies - the revert was intentional, but I accidentally added it to another commit it seems. The build seemed to be failing with that version, and I didn't have to to investigate the cause, so I postponed that version bump until later.

ack, will not bump that yet! i think upstream changed how things are orgnized so it could be related to that.

Separately, trying to figure why the build fails in github ci, but works outside of it.

jocado (commented):
@xnox    It looks like we are dropping the requirement for https://github.com/docker-snap/docker-snap/blob/main/patches/libnvidia-container/nvidia-0001-Fix-for-22.04-build.patch , so perhaps rm the patch file too ?

jocado:
> Separately, trying to figure why the build fails in github ci, but works outside of it.

Just having a look now too..

jocado:
Not massively helpful - but I can confirm at least that the build works for me locally too, using the multipass snapcraft provider.

xnox:
@jocado my patches by accident droped `patch.sh` step for libnvidia part, which is what was causing my ci to fail. now fixed i think. So should work here finally.... i hope.....

jocado:
Ahh, thanks @xnox , makes sense, I was just looking at that patch.

I presumed earlier when I saw the PR that you have dropped it on purpose, hence my comment above :) 
https://github.com/docker-snap/docker-snap/pull/131#pullrequestreview-1428321517

PR #134, opened 2023-05-17
Assumes did nothing, because snapcraft.

xnox:


xnox:
Also see https://github.com/snapcore/pc-gadget/pull/92

PR #141, opened 2023-05-24
Add simple smoke-test to ensure init is supported

jocado:
This adds a simple test to check for init support, and prevent the issue: https://github.com/docker-snap/docker-snap/issues/139


jocado:
@xnox Here is a test which covers the feature mentioned here: https://github.com/docker-snap/docker-snap/pull/140

PR #144, opened 2023-06-05
Improvements to snap refresh related to nvidia support 

jocado:
This addresses two current known issues to snap refresh if nvidia support is in use.

The previous CDI config is never valid for a newer revision, as some paths contain the snap revision.
- This PR ensures the previous config is always purged

If the `nvidia-container-toolkit` service fails to start properly for some reason, it prevents the snap refresh from completing, even though dockerd would remain functional for non nvidia images use cases.
- This PR enables `nvidia-container-toolkit` to fail setting up the CDI and dockerd config, and print a warning instead.
- It also ensures that dockerd config is only attempted if CDI config is generated.
- CDI config can fail on refresh due to: `ERROR_LIB_RM_VERSION_MISMATCH` [ nvidia kernel module and user space libs do not match ].

### DIFF
--- a/bin/nvidia-container-toolkit
+++ b/bin/nvidia-container-toolkit
@@ -28,7 +28,7 @@ device_wait() {
 
 cdi_generate () {
     # Generate the CDI config #
-    "${SNAP}/usr/bin/nvidia-ctk" cdi generate --nvidia-ctk-path "${SNAP}/usr/bin/nvidia-ctk" --output="${SNAP_DATA}/etc/cdi/nvidia.yaml"
+    "${SNAP}/usr/bin/nvidia-ctk" cdi generate --nvidia-ctk-path "/snap/${SNAP_NAME}/current/usr/bin/nvidia-ctk" --output="${SNAP_DATA}/etc/cdi/nvidia.yaml"
 }
 
 runtime_configure () {
--- a/bin/nvidia-container-toolkit
+++ b/bin/nvidia-container-toolkit
@@ -36,8 +36,17 @@ runtime_configure () {
     "${SNAP}/usr/bin/nvidia-ctk" runtime configure --runtime=docker --runtime-path "${SNAP}/usr/bin/nvidia-ctk" --config "${SNAP_DATA}/config/daemon.json"
 }
 
-setup_warn () {
+setup_fail () {
     echo "WARNING: Conainter Toolkit setup seemed to fail with an error"
+
+    # Remove nvidia runtime config, if it exists #
+    jq -r 'del(.runtimes.nvidia)' "${SNAP_DATA}/config/daemon.json" > "${SNAP_DATA}/config/daemon.json.new"
+
+    # If it was removed [ there was a change ], copy in the new config, remove CDI config,  and set service restart flag #
+    if ! cmp "${SNAP_DATA}/config/daemon.json"{,.new} >/dev/null ; then
+        mv "${SNAP_DATA}/config/daemon.json"{.new,}
+        rm -f "${SNAP_DATA}/etc/cdi/nvidia.yaml"
+    fi
 }
 
 setup_info () {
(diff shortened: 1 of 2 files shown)


jocado:
@xnox PTAL when you get a chance - thanks.

bin/nvidia-container-toolkit
```diff
@@ -26,6 +26,33 @@ device_wait() {
...
+    if ! cmp "${SNAP_DATA}/config/daemon.json"{,.new} >/dev/null ; then
+        mv "${SNAP_DATA}/config/daemon.json"{.new,}
+        rm -f "${SNAP_DATA}/etc/cdi/nvidia.yaml"
+    fi
+}
+
+setup_info () {
+    echo "Conainter Toolkit setup complete"
+}
+
```
xnox, line 55: I wonder if nvidia support should be disabled with snap set upon failing to create it. As likely kernel & toolkit libraries refresh would be required to resolve it.

snap/hooks/post-refresh
```diff
@@ -5,3 +5,6 @@ if [ ! -f "$SNAP_DATA/config/daemon.json" ]; then
     mkdir -p "$SNAP_DATA/config"
     cp "$SNAP/config/daemon.json" "$SNAP_DATA/config/daemon.json"
 fi
+
+# Should should always be re-craeted if required, as it contains paths based on snap revision #
+rm -f "${SNAP_DATA}/etc/cdi/nvidia.yaml"
```
xnox, line 10: snap revisions of what? docker itself? but SNAP_DATA is per-snap-revision directory already. Or does it get copied up between refreshes?

jocado: An example path from the CDI config:
```
/snap/docker/2893/usr/bin/nvidia-ctk
```

As the CDI config file is in `$SNAP_DATA`, it get's copied on snap update.  It does need to be generated in-situ on the host.

I'm not aware of a better way to handle this scenario, and this seems reasonable.

bin/nvidia-container-toolkit
jocado: >As likely kernel & toolkit libraries refresh would be required to resolve it.

@xnox 

That's true.  It would then also require specific action to re-enable however.

It's hard to try and anticipate all different end user scenarios, but I think on balance not blocking install is best.  In the scenarios where non-nvidia containers are also required this doesn't block that.

The only other option I can think of is to make this behaviour configurable.  Something like:
```
nvidia-support.strict: true|false
```
- `true`
  - install fails if nvidia support fails to be configured
- `false`
  - install continues if nvidia support fails to be configured

What do do think ?

snap/hooks/post-refresh
xnox: The rest of the configs are in $SNAP_DATA, and you would be able to do it in snap data too.... for example I would expect it to use `/snap/docker/current/usr/bin/nvidia-ctk` or even just use `nvidia-ctk` as that path should be in `$PATH` and just work.

I'll play with this, but I don't like generating/regenerating files with snap-internal paths, when it really ought to be docker revision agnostic, especially when none of the other nvidia things change.

jocado: That path used for that command is specified when we generate the config.  We can change it from:
```
"${SNAP}/usr/bin/nvidia-ctk" cdi generate --nvidia-ctk-path "${SNAP}/usr/bin/nvidia-ctk" --output="${SNAP_DATA}/etc/cdi/nvidia.yaml"
```
to
```
"${SNAP}/usr/bin/nvidia-ctk" cdi generate --nvidia-ctk-path "/snap/SNAP_NAME/current/usr/bin/nvidia-ctk" --output="${SNAP_DATA}/etc/cdi/nvidia.yaml"
```
I will update and remove the purge.

I tested no fully qualified path, but the canonical path is then discovered and used [ include snap revision ].

jocado: Updated - PTAL

xnox (approved):

PR #150, opened 2023-07-27
Backport patches to enable support for changes to latest golang

jocado:
Without this patch, snap builds produce a snap with a non-functional docker cli and client, they fail with `http: invalid Host header`

REF: https://github.com/moby/moby/issues/45935

### DIFF
--- a/patches/docker-cli/local-connections-dummy-hostname.patch
+++ b/patches/docker-cli/local-connections-dummy-hostname.patch
@@ -1,0 +1,83 @@
+diff --git a/vendor/github.com/docker/docker/client/client.go b/vendor/github.com/docker/docker/client/client.go
+index 9b2b2eaeb8..5bf699d88d 100644
+--- a/vendor/github.com/docker/docker/client/client.go
++++ b/vendor/github.com/docker/docker/client/client.go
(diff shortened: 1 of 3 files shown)


jocado:
For ref, here is an example failure for an action log: https://github.com/docker-snap/docker-snap/actions/runs/5684149041/job/15406223201#step:10:6

lucaskanashiro:
Thanks for this PR @jocado ! I see the patch you are proposing was already merged upstream:

https://github.com/moby/moby/commit/dab9ffb252186f4c47416addb136d579f9314c6a

This fix is present in version `24.0.5`:

https://github.com/moby/moby/releases/tag/v24.0.5

I am planning to push `24.0.5` to `latest/edge` soon but it would take a while until we promote it to `latest/stable`. So I think the best route is to merge your change, follow all the process to promote it to `latest/stable` and then update `latest/edge` to `24.0.5`.

The smoke test also passed against your branch, so I believe we are good to go.

lucaskanashiro:
If you want to test yourself, I just uploaded it to the snap store:

`Revision 2903 created for 'docker' and released to 'latest/edge'`

PR #152, opened 2023-08-24
Import 24.x

lucaskanashiro:


xnox (approved):
looks good.

xnox:
@lucaskanashiro looks like github workflows are broken, so ideally those need to be fixed up.

lucaskanashiro:
> @lucaskanashiro looks like github workflows are broken, so ideally those need to be fixed up.

I am going to take a look at it. In a first glance, it seems to be an issue in the Snapcraft build action (when installing LXD and other dependencies), I'll try to understand better what's happening.

PR #153, opened 2023-09-06
Nvidia runtime default

jocado:
Fix bug in runtime config
    
Should be nvidia-container-runtime binary, not nvidia-ctk
    
NVIDIA GPU support works without using the full nvidia-container-runtime,
but in some cases it turns out that switching to the
nvidia-container-runtime entirely is beneficial [ ability to schedule multiple simultaneous GPU containers ]

Example usage:
```
docker run --rm --runtime=nvidia -e NVIDIA_VISIBLE_DEVICES=all {image-name} {container-name}
```

### DIFF
--- a/snap/hooks/post-refresh
+++ b/snap/hooks/post-refresh
@@ -1,5 +1,9 @@
 #!/bin/bash -e
 
+# ensure the layouts dir for /etc/{stuff} exists
+mkdir -p "$SNAP_DATA/etc/docker"
+mkdir -p "$SNAP_DATA/etc/nvidia-container-runtime"
+
 # copy the config file from $SNAP into $SNAP_DATA if it doesn't exist
 if [ ! -f "$SNAP_DATA/config/daemon.json" ]; then
     mkdir -p "$SNAP_DATA/config"
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -115,8 +115,7 @@ apps:
       - privileged
       - support
       - graphics-core22
-      # FIXME: check `system-observe` is really required
-      - system-observe
+      - system-observe # Seems funtional without - but it generates quite a bit of noise in audit logs - can be manually connected
     slots:
       - docker-daemon
 
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -46,6 +46,8 @@ layout:
   # Container Device Interface (CDI) Support - https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html#container-device-interface-cdi-support
   /etc/cdi:
     bind: $SNAP_DATA/etc/cdi
+  /etc/nvidia-container-runtime:
+    bind: $SNAP_DATA/etc/nvidia-container-runtime
   /etc/gitconfig:
     bind-file: $SNAP_DATA/etc/gitconfig
   /usr/libexec/docker/cli-plugins:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -115,6 +115,7 @@ apps:
       - privileged
       - support
       - graphics-core22
+      # FIXME: check `system-observe` is really required
       - system-observe
     slots:
       - docker-daemon
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -115,7 +115,6 @@ apps:
       - privileged
       - support
       - graphics-core22
-      - system-observe # Seems funtional without - but it generates quite a bit of noise in audit logs - can be manually connected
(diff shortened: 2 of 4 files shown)


jocado:
> Hi @lucaskanashiro
> 
> Would it be possible to merge this change [ which essentially fixes a bug ] so it could be bumped to stable before the changes in #152 ?
> 
> Thanks!

Hi @lucaskanashiro - just wondering if you saw this request and had any feedback ?

Thank you.

snap/hooks/post-refresh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash -e
 
+# ensure the layouts dir for /etc/{stuff} exists
+mkdir -p "$SNAP_DATA/etc/docker"
```
lucaskanashiro, line 4: In theory we shouldn't need to make sure that the `/etc/docker` directory exists, since it is created in the install hook for many years (since 2018), so I bet that all current installations should have that in place. On the other side, this is harmless.

jocado: I agree it very likely not required for most scenarios in post-refresh, especially as it's been in place in inst for while.

Really, I just did to sync between the two hooks, because I think it makes more sense to just sync that rather than try and maintain it over time.  e.g.  I added the `mkdir` for `nvidia-container-runtime` in install and post-refresh because it was new, and then after _some_ arbitrary period you could argue that it could be removed from post-refresh, but perhaps it's just simpler to leave it synced.

I think it's ok, but I'm happy to remove it if you prefer.

lucaskanashiro: I would be fine in adding this to post-refresh.

snap/snapcraft.yaml
```diff
@@ -113,6 +115,7 @@ apps:
       - privileged
       - support
       - graphics-core22
+      - system-observe # Seems funtional without - but it generates quite a bit of noise in audit logs - can be manually connected
```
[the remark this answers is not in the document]
jocado: Unfortunately, as some time has passed, I have forgetten the full reason.  When I tried to replicate it just now, I the remaining audit lines I saw weren't too much.  I'm happy to remove, and re-visit if I found the proper reason again.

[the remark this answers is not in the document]
jocado: Updated: https://github.com/docker-snap/docker-snap/pull/153/commits/77e990a39465016c3ade9e57a932c045a24122d3

snap/hooks/post-refresh
jocado: Cool - let's leave then :+1:

lucaskanashiro:
OK. The changes now look good to me. Let's wait for the CI results to approve this.

jocado:
Great - thanks for review :+1: 

Do you have any rough idea about timescales for promotion through channels ?

lucaskanashiro:
It will depend on the internal testing which might take some weeks (based on the previous revision). BTW, The last PR I merged from you is already in the candidate channel, feel free to test it out.

jocado:
Sure - I already tested it - thank you :smile: 

I will keep an eye on the revisions over the next few weeks.

lucaskanashiro:
I just merged it but now I noticed we should have squashed some commits to keep the history clean. Let's try to do it next time.

jocado:
I will try and remember that for next time - thanks for your review and help :+1:

YamiYukiSenpai:
```
$ sudo docker run -d --name jellyfin --net=host --volume /home/.jellyfin/docker/config:/config --volume /home/.jellyfin/docker/cache:/cache --mount type=bind,source=/media,destination=/media,ro=false --user 1001:1001 --device /dev/nvidia0:/dev/nvidia0 --device /dev/nvidiactl:/dev/nvidiactl --device /dev/nvidia-modeset:/dev/nvidia-modeset --device /dev/nvidia-uvm:/dev/nvidia-uvm --device /dev/nvidia-uvm-tools:/dev/nvidia-uvm-tools --runtime=nvidia --gpus all jellyfin/jellyfin
docker: Error response from daemon: unknown or invalid runtime name: nvidia.

```

Still no dice for me

jocado:
Hi @YamiYukiSenpai 

Couple of things:
- A closed MR is not a great place to ask for help or feedback about issues - please use the project Issues for this.
- You don't specify what you OS it, but as noted in the [README](https://github.com/docker-snap/docker-snap#nvidia-support-on-ubuntu-core-22) NVIDIA GPU support is only available on Ubuntu Core 22 currently, not Ubuntu Classic, please see these already open issues:
  - https://github.com/docker-snap/docker-snap/issues/127
  - https://github.com/docker-snap/docker-snap/issues/148

PR #158, opened 2024-01-18
Provide content slots for ro/rw access to docker daemon config

alexclewontin:
There are use cases where another snap may want to either observe or modify the contents of daemon.json. One such case is azure-iot-edge, which does some networking diagnostics based on settings it reads from that file. It only needs read access, but it is also totally feasible that unattended IoT deployments may want to configure their daemon.json in certain ways and writing via a content interface would be one of the easier ways to do that.

No strong opinions on the naming, or whether these should be collapsed into a single R/W interface (as opposed to also having the R/O version)

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -80,6 +80,16 @@ plugs:
     # Currently, this is the only relevant provider, but cannot be default as not supported on all archs #
     #default-provider: nvidia-core22
 slots:
+  config:
+    interface: content
+    content: docker-config
+    write:
+      - $SNAP_DATA/config
+  config-ro:
+    interface: content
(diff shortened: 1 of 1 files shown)


alexclewontin:
> ro seems safe
> 
> I'm not sure about RW => how would that work together with `snap set` and the .json being written out during snap upgrades; restarts; settings changes?
> 
> Unless we only do it once on init, and never change again. I can't remember anymore.
> 
> For example, after RW changes, who will be responsible to restart the daemon at the very least? rollback if changes are bogus?

all fair Qs. Since my current use case is really just the r/o case, I'll drop the r/w case for the moment, and we can re-address when someone has a clear need

xnox (approved):

PR #165, opened 2024-07-08
Add Nvidia workflow and basic tests

locnnil:
Add workflow to test docker snap with Nvidia GPUs.
These tests depend on hardware with GPUs, available in Canonical's test farm.

The workflow adds a manual trigger which takes the snap branch as input. This can be used to test a pre-built installed from a channel. 

In future, the existing smoke test workflow should change to upload the built snap to the store when triggered from a PR, to in a branch, for example `latest/edge/pr-xx`. Then, the nvidia-test workflow can be triggered to test the snap.

The reason for keeping the workflows separate is to run the nvidia-tests only when necessary and by maintainers. This is primarily to save resources.


farshidtz:
Approve /canonical/self-hosted-runners/run-workflows da2c111250fa67750992e3cfd43b99ede5791269

.github/workflows/nvidia-test.yml
```diff
@@ -0,0 +1,70 @@
...
+                snapcraft-args: snap --output docker.snap
+            
+            - name: Install snapcraft
+              run: |
+                sudo snap install snapcraft --classic
+            
+            - name: Upload artifact to snapstore branch
+              run: |
+                export SNAPCRAFT_STORE_CREDENTIALS="${{ secrets.SNAPCRAFT_LOGIN }}"
+                snapcraft upload ${{steps.snapcraft.outputs.snap}} --release latest/edge/pr-${{ env.PR_NUMBER }}
```
locnnil, line 42: @lucaskanashiro the plan is to have this secret to login in the CI machine and upload to a temporary branch.

.github/testflinger/nvidia.yaml
farshidtz, line 1: This is obsolete.

.github/workflows/nvidia-test.yml
```diff
@@ -0,0 +1,54 @@
+name: Nvidia Graphics Tests
+
+on:
+  # Uncomment this trigger only during workflow development
+  pull_request:
+    branches: [ main ]
```
farshidtz, line 6: Comment before merging.

lucaskanashiro (approved):
I did not test it locally but the code changes LGTM. Thanks for your work on that!

PR #170, opened 2024-08-21
Change GH workflows to publish the snap to Store for Nvidia testing

locnnil:
Add possibility to run nvidia workflow for artifacts of PRs and separate build and upload artifact on a dedicated workflow.

- Add: workflow to build and upload artifact
- Feat: Add possibility to run nvidia tests by passing the run-id of the wanted build workflow.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,38 @@
+name: Build
+
+on:
+  pull_request:
+  push:
+
+on:
+  workflow_run:
+    workflows: ["trigger"]
+    types:
+      - completed
+
+defaults:
+  run:
+    shell: 'bash -Eeuo pipefail -x {0}'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Build
+        uses: snapcore/action-build@v1
+        id: snapcraft
+        with:
+          snapcraft-args: snap --output docker_${{ github.run_id}}.snap
+
+      - name: Upload
+        # TODO: the goal of this "if:" condition is to avoid creating artifacts on private forks (and using up all their storage quota),
+        # but Tianon couldn't find a clean way to do that so this was the next best thing
+        if: github.repository == 'canonical/docker-snap'
+        uses: actions/upload-artifact@v4
+        with:
+          name: docker_${{ github.run_id}}.snap
+          path: ${{ steps.snapcraft.outputs.snap }}
+
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,12 +4,6 @@ on:
   pull_request:
   push:
 
-on:
-  workflow_run:
-    workflows: ["trigger"]
-    types:
-      - completed
-
 defaults:
   run:
     shell: 'bash -Eeuo pipefail -x {0}'
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,10 +4,6 @@ on:
   pull_request:
   push:
 
-defaults:
-  run:
-    shell: 'bash -Eeuo pipefail -x {0}'
-
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -22,9 +18,6 @@ jobs:
           snapcraft-args: snap --output docker_${{ github.run_id}}.snap
 
       - name: Upload
-        # TODO: the goal of this "if:" condition is to avoid creating artifacts on private forks (and using up all their storage quota),
-        # but Tianon couldn't find a clean way to do that so this was the next best thing
-        if: github.repository == 'canonical/docker-snap'
         uses: actions/upload-artifact@v4
         with:
           name: docker_${{ github.run_id}}.snap
--- a/.github/workflows/nvidia-test.yml
+++ b/.github/workflows/nvidia-test.yml
@@ -34,15 +34,15 @@ jobs:
               uses: dawidd6/action-download-artifact@v6
               with:
                 workflow: build.yml
-                name: docker_${{ github.event.inputs.run_id }}.snap
+                name: docker_${{ inputs.run_id }}.snap
 
             - name: Publish to Store
               uses: snapcore/action-publish@v1
               env:
                 SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_LOGIN }}
               with:
-                snap: docker_${{ github.event.inputs.run_id }}.snap
-                release: latest/edge/runid-${{ github.event.inputs.run_id }}
+                snap: docker_${{ inputs.run_id }}.snap
+                release: latest/edge/runid-${{ inputs.run_id }}
 
             - name: Create Testflinger job queue
               run: |
--- a/.github/workflows/nvidia-test.yml
+++ b/.github/workflows/nvidia-test.yml
@@ -47,7 +47,7 @@ jobs:
             - name: Create Testflinger job queue
               run: |
                 export JOB_QUEUE="${{ matrix.job-queue }}"
-                export SNAP_CHANNEL="${{ inputs.snap-channel }}"
+                export SNAP_CHANNEL="latest/edge/runid-${{ github.event.inputs.run_id }}"
 
                 envsubst '$JOB_QUEUE' \
                   < $TESTFLINGER_DIR/nvidia-job.yaml \
--- a/.github/workflows/nvidia-test.yml
+++ b/.github/workflows/nvidia-test.yml
@@ -7,11 +7,10 @@ on:
   # Manual trigger
   workflow_dispatch:
     inputs:
-      snap-channel:
-        description: 'Docker snap channel'
+      run_id:
+        description: 'Run id number'
         required: true
-        type: string
-        default: 'latest/edge'
+        type: number
 
 jobs:
     test:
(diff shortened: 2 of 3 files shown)


hartmutobendorf:
sorry for causing confusion, the reference was meant to be to an issue, not this pr

locnnil:
I suspect that the smoke-tests workflow isn’t being triggered because the “Build” workflow doesn’t exist in main branch.

.github/workflows/smoke-test.yml
```diff
@@ -48,7 +44,8 @@ jobs:
 
       - name: Install
         run: |
-          sudo snap install --dangerous ./docker.snap
+          cd ./docker_*.snap
```
locnnil, line 47: Here it's a directory because as I'm using a regex for the name, the action `dawidd6/action-download-artifact` creates a repository.

.github/workflows/smoke-test.yml
```diff
@@ -48,7 +45,8 @@ jobs:
...
```
farshidtz, line 48:
```suggestion
          # Change to the directory where the snap has been downloaded by the artifact download action 
          cd ./docker_*.snap
```

.github/workflows/smoke-test.yml
farshidtz: Please add this as comments in the code, so future readers also be aware of it: https://github.com/canonical/docker-snap/pull/170#discussion_r1730016927

.github/workflows/build.yaml
```diff
@@ -0,0 +1,32 @@
+name: Build
+
+on:
+  pull_request:
+  push:
+
+defaults:
+  run:
+    shell: 'bash -Eeuo pipefail -x {0}'
```
farshidtz, line 9:
```suggestion
```
There are no jobs/steps using shell in this workflow.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Build
+        uses: snapcore/action-build@v1
+        id: snapcraft
+        with:
+          snapcraft-args: snap --output docker_${{ github.run_id}}.snap
+
+      - name: Upload
+        # TODO: the goal of this "if:" condition is to avoid creating artifacts on private forks (and using up all their storage quota),
+        # but Tianon couldn't find a clean way to do that so this was the next best thing
+        if: github.repository == 'canonical/docker-snap'
```
farshidtz, line 27:
```suggestion
```
If the artifact isn't uploaded on forks, the smoke test would fail when trying to download it.

.github/workflows/nvidia-test.yml
```diff
@@ -31,11 +30,25 @@ jobs:
...
+                snap: docker_${{ github.event.inputs.run_id }}.snap
+                release: latest/edge/runid-${{ github.event.inputs.run_id }}
+
             - name: Create Testflinger job queue
               run: |
                 export JOB_QUEUE="${{ matrix.job-queue }}"
-                export SNAP_CHANNEL="${{ inputs.snap-channel }}"
-                
+                export SNAP_CHANNEL="latest/edge/runid-${{ github.event.inputs.run_id }}"
+
```
farshidtz, line 51: All instances of `github.event.inputs.run_id` can be simplified to `inputs.run_id`, as in the removed code on line 37. I see this the former mentioned in a [blog post from 2020](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/), but official docs reference the [manual trigger](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow) inputs as latter.

.github/workflows/smoke-test.yml
```diff
@@ -48,7 +45,9 @@ jobs:
 
       - name: Install
         run: |
-          sudo snap install --dangerous ./docker.snap
+          # dawidd6/action-download-artifact action creates a directory when the name is fetch as a RegExp
```
farshidtz, line 48:
```suggestion
          # dawidd6/action-download-artifact action creates a directory when the name is given as a regex
```

.github/workflows/smoke-test.yml
```diff
@@ -48,7 +45,9 @@ jobs:
...
+          cd ./docker_*.snap
+          sudo snap install --dangerous ./docker_*.snap
           # it will immediately be failing to start, so let's head that off as quickly as we can
```
farshidtz, line 51:
```suggestion
          sudo snap install --dangerous ./docker_*.snap
          
          # it will immediately be failing to start, so let's head that off as quickly as we can
```

.github/workflows/smoke-test.yml
farshidtz: To separate workflow logic from the test's set up, you could change the working directory from the step itself, moving the above comment there.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsworking-directory

.github/workflows/smoke-test.yml
```diff
@@ -12,22 +14,17 @@ jobs:
   smoke-test:
     name: Snapcraft
```
farshidtz, line 15: Please remove misleading job name from existing code.

farshidtz (changes requested):

.github/workflows/build.yaml
locnnil: Fixed! Thanks!

.github/workflows/build.yaml
locnnil: Yeah, that's true, done!

.github/workflows/nvidia-test.yml
locnnil: Done!

.github/workflows/smoke-test.yml
locnnil: Done!

.github/workflows/smoke-test.yml
locnnil: I'm using "RegExp" here to match the name of the option `name_is_regexp`, but I replace "fetch" -> "given"

The `working-directory` instruction doesn't accept wildcards, [I've just tested it here](https://github.com/locnnil/ci-exp/actions/runs/10549822156/job/29225312984#step:4:7).

.github/workflows/nvidia-test.yml
farshidtz: Thanks but this isn't done. Only one instance has been changed.

locnnil: True! Fixed the other on https://github.com/canonical/docker-snap/pull/170/commits/1a51e01f4f059f88f998c3b21746fccd066916c8

farshidtz: THERE ARE TWO MORE!

farshidtz (approved):
Looks good. Thanks. Need to merge to test the workflow triggers.

farshidtz:
The smoke test workflow fails during the artifact download step. 
https://github.com/canonical/docker-snap/actions/runs/10555183107/job/29238305965

```
Run dawidd6/action-download-artifact@v6
  with:
    workflow: build.yml
    name: docker_[0-9]+.snap
    name_is_regexp: true
    github_token: ***
    workflow_search: false
    workflow_conclusion: success
    repo: canonical/docker-snap
    path: ./
    allow_forks: false
    check_artifacts: false
    search_artifacts: false
    skip_unpack: false
    if_no_artifact_found: fail
==> Repository: canonical/docker-snap
==> Artifact name: docker_[0-[9](https://github.com/canonical/docker-snap/actions/runs/10555183107/job/29238305965#step:3:9)]+.snap
==> Local path: ./
==> Workflow name: build.yml
==> Workflow conclusion: success
==> Allow forks: false
Error: Not Found - https://docs.github.com/rest/actions/workflow-runs#list-workflow-runs-for-a-workflow
```

Looks like it is can't find the workflow because the given workflow name (build.yml) doesn't match the actual name (build.yaml).

.github/workflows/smoke-test.yml
```diff
@@ -1,33 +1,29 @@
...
-          snapcraft-args: snap --output docker.snap
-
-      - name: Upload
-        # TODO the goal of this "if:" condition is to avoid creating artifacts on private forks (and using up all their storage quota), but Tianon couldn't find a clean way to do that so this was the next best thing
-        if: github.repository == 'docker-snap/docker-snap'
-        uses: actions/upload-artifact@v2
-        with:
-          name: docker.snap
-          path: docker.snap
+          workflow: build.yml
```
farshidtz, line 24: This is actually very fragile. Can it be replaced with the workflow name, matching the name given in the trigger (`"Build"`)?

From the docs:
```
   # Optional, workflow file name or ID
   # If not specified, will be inferred from run_id (if run_id is specified), or will be the current workflow
   workflow: workflow_name.yml
```
Looking at the action's source [here](https://github.com/dawidd6/action-download-artifact/blob/db9477a3ebd885bf3bb8da6a98eeb0d6da2d0d9a/main.js#L60-L66), it seems that by workflow ID, they mean its name.

.github/workflows/nvidia-test.yml
```diff
@@ -31,11 +30,25 @@ jobs:
             - name: Checkout code
               uses: actions/checkout@v4
 
+            - name: Get the artifact
+              uses: dawidd6/action-download-artifact@v6
+              with:
+                workflow: build.yml
```
farshidtz, line 36: Can this be replaced with `run_id: ${{ inputs.run_id }}` to make the relation between workflows less fragile?

farshidtz: Actually, how can the action know which build run to look for based on just the workflow file name?

In [this case](https://github.com/canonical/docker-snap/actions/runs/10557474952/job/29245052143), the Nvidia workflow can't find the right artifact:
```
Run dawidd6/action-download-artifact@v6
==> Repository: canonical/docker-snap
==> Artifact name: docker_10555346598.snap
==> Local path: ./
==> Workflow name: build.yml
==> Workflow conclusion: success
==> Allow forks: false
==> (found) Run ID: 105573532[18](https://github.com/canonical/docker-snap/actions/runs/10557474952/job/29245052143#step:4:19)
==> (found) Run date: 2024-08-26T09:33:57Z
==> (not found) Artifact: docker_10555346598.snap
==> Found the following artifacts instead:
	==> (found) Artifact: docker_10557353[21](https://github.com/canonical/docker-snap/actions/runs/10557474952/job/29245052143#step:4:22)8.snap
```

Replacing that with run id should resolve this.

locnnil: Added here: https://github.com/canonical/docker-snap/pull/174/files

PR #171, opened 2024-08-21
fix: allow execution of /usr/bin/ssh

### LINKED ISSUE #169: snap can not run /usr/bin/ssh => docker cannot use remote context
### Why this is important
When running docker in a VM, or on a production host, I cannot manage the containers from my local machine.
**This also means I cannot use/access containers in my IDE (VS Code).** Running a docker blueprint in multipass or lxd is only possible with an install of the upstream, unsnapped docker.

### To reproduce:
``` bash
docker context create rem --docker "host=ssh://user@host"
docker context use rem
docker ps
```
fails with the error message
`error during connect: Get "http://docker.example.com/v1.24/containers/json": fork/exec /usr/bin/ssh: permission denied`

hartmutobendorf:
Closes https://github.com/canonical/docker-snap/issues/169

added ssh-keys interface
added ssh-public-keys interface
added openssh-client for staging
added bind-file layout for /use/bin/ssh


locnnil (changes requested):
Also, please, remember to:

-  Update the documents mentioning this new interfaces and giving the right orientation on when the user might want to use then.
- Add a new test for this functionality.

snap/snapcraft.yaml
```diff
@@ -151,6 +155,11 @@ parts:
       - jq
       - pciutils
 
+  ssh:
+    plugin: nil
+    stage-packages:
+      - openssh-client
```
locnnil: To enable password authentication:
```suggestion
      - openssh-client
      - ssh-askpass
```

locnnil (commented):
Changes looks good! I have one sidenote and one important remark on this:

Sidenote: if you have a non-standard name for your ssh keys (anything other than `~/.ssh/id_rsa` and `~/.ssh/id_rsa.pub`) you should set the identity file to `~/.ssh/config`:
```config
Host <host>
    User <user>
    IdentityFile ~/.ssh/<my_key>
```

Otherwise, it will fail with:
```console
$ docker --context rem ps
error during connect: Get "http://docker.example.com/v1.52/containers/json": command [ssh -l pi -o ConnectTimeout=30 -T -- <ip-address> docker system dial-stdio] has exited with exit status 255, make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=ssh_askpass: exec(/usr/bin/ssh-askpass): No such file or directory
Permission denied, please try again.
ssh_askpass: exec(/usr/bin/ssh-askpass): No such file or directory
Permission denied, please try again.
ssh_askpass: exec(/usr/bin/ssh-askpass): No such file or directory
pi@<ip-address>: Permission denied (publickey,password).
```
But I think it would be important to also support password authentication over ssh, so I left a suggestion on adding the `ssh-askpass` package.  

--- 

The important remark: snapd appends `/snap/bin` to the `$PATH` on `/etc/profile.d/apps-bin-path.sh`.

But over ssh, this script isn't sourced, so the `$PATH` doesn't contain the `/snap/bin` directory: 
```console
ssh pi@buri.local printenv
SHELL=/bin/bash
PWD=/home/pi
LOGNAME=pi
XDG_SESSION_TYPE=tty
MOTD_SHOWN=pam
HOME=/home/pi
LANG=en_GB.UTF-8
SSH_CONNECTION=192.168.68.109 60178 192.168.68.105 22
XDG_SESSION_CLASS=user
USER=pi
SHLVL=0
XDG_SESSION_ID=186
XDG_RUNTIME_DIR=/run/user/1000
SSH_CLIENT=192.168.68.109 60178 22
PATH=/usr/local/bin:/usr/bin:/bin:/usr/games
DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus
_=/usr/bin/printenv
```

There are some workarounds to this like:

- Option 1:
You can force the PATH to contain `/snap/bin` on the remote by setting it in [environment.d(5)](https://man7.org/linux/man-pages/man5/environment.d.5.html) on `/etc/environment`, but it's necessary to set the entire path and not append it, like:
```console
$ cat /etc/environment
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/games:/usr/games:/snap/bin
```

- Option 2:
You can register the docker snap path as an alternative on the remote:
```shell
sudo update-alternatives --install /usr/bin/docker docker /snap/bin/docker 50
```
**WARNING**: This would definitely conflict if you have docker installed as a deb.

Currently, we list all the available interfaces on our `README.md`, so please update the list of interfaces and add a note on the usage of the ssh interfaces.

PR #172, opened 2024-08-26
ci: Rename build.yaml to build.yml

farshidtz:
The build workflow has been added in https://github.com/canonical/docker-snap/pull/170

But the file name doesn't match the name given in the download steps in smoke and nvidia tests.

Renaming to yml because this makes the extension consistent with other workflows in this repo.


locnnil (approved):
LGTM! +1!

PR #174, opened 2024-08-26
ci: Make the artifact download workflow step less dependent on build workflow name

locnnil:
Remove `workflow` optional argument and set `workflow_search` to true, so the artifact can be matched based on the other given characteristics.

### DIFF
--- a/.github/workflows/nvidia-test.yml
+++ b/.github/workflows/nvidia-test.yml
@@ -33,7 +33,8 @@ jobs:
             - name: Get the artifact
               uses: dawidd6/action-download-artifact@v6
               with:
-                workflow: build.yml
+                workflow_search: true
+                run_id: ${{ inputs.run_id }}
                 name: docker_${{ inputs.run_id }}.snap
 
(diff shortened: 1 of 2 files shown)


.github/workflows/nvidia-test.yml
```diff
@@ -33,7 +33,8 @@ jobs:
             - name: Get the artifact
               uses: dawidd6/action-download-artifact@v6
               with:
-                workflow: build.yml
+                workflow_search: true
```
farshidtz, line 36:
```suggestion
```
Is a search needed when the action receives the unique run id?

https://github.com/dawidd6/action-download-artifact/blob/db9477a3ebd885bf3bb8da6a98eeb0d6da2d0d9a/main.js#L61

farshidtz (approved):
LGTM

Resolving to https://github.com/canonical/docker-snap/pull/170#discussion_r1730829354 and https://github.com/canonical/docker-snap/pull/170#discussion_r1730838394

PR #175, opened 2024-08-26
ci: merge build and smoke-test into one workflow

locnnil:
With the changes added in #170, the smoke test results no longer show up on PRs. This is because the smoke tests trigger by the build workflow's completion, rather than the PR.

Merging the workflows into one, so that the result of both build and smoke tests can be integrated into PR checks.


farshidtz (commented):
Replaced the PR description to explain the reason for this change.

farshidtz (changes requested):
Smoke test job is being skipped.

Need to remove this obsolete condition from the job: `if: ${{ github.event.workflow_run.conclusion == 'success' }}`

PR #178, opened 2024-08-30
Update to version 27.2.0

### LINKED ISSUE #167: CVE-2024-41110
Hello.

Are there already efforts to release a new version that fixes the problem? 

https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2024-41110
https://github.com/moby/moby/security/advisories/GHSA-v23v-6jw2-98fq
https://docs.docker.com/engine/extend/plugins_authorization/

Docker has fixed it with version 27.1.1 (2024-07-23)
https://docs.docker.com/engine/release-notes/27.1/

locnnil:
- Rewrote the `snappy-apparmor-tweaks` patch to ensure it applies correctly.
- Rewrote the `snappy-buildkit-git-environ` patch to ensure it applies correctly.
- Updated Go from version **1.20** to **1.21**.
- Updated Moby from version **v24.0.5** to **v27.2.0**.
- Updated containerd from version **v1.6.21** to **v1.7.21**.
- Updated runc from version **v1.1.12** to **v1.1.13.**
- closes #167.

### DIFF
--- a/patches/engine/0002-snappy-apparmor-tweaks.patch
+++ b/patches/engine/0002-snappy-apparmor-tweaks.patch
@@ -17,7 +17,7 @@ index 277c853ebe..92e3872722 100644
  	"os"
  	"os/exec"
  	"path"
-+  "strconv"
++	"strconv"
  	"strings"
  	"text/template"
  )
--- a/patches/engine/0002-snappy-apparmor-tweaks.patch
+++ b/patches/engine/0002-snappy-apparmor-tweaks.patch
@@ -1,28 +1,27 @@
-From e9302d52e7cdc7b4b6364f08dfe15d52b920fff5 Mon Sep 17 00:00:00 2001
-From: Lucas Kanashiro <lucas.kanashiro@canonical.com>
-Date: Wed, 23 Aug 2023 18:58:55 -0300
+From 173566eebc60e3ed18ec595686688ad10f25388b Mon Sep 17 00:00:00 2001
+From: Lincoln Wallace <lincoln.wallace@canonical.com>
+Date: Wed, 28 Aug 2024 17:30:12 -0300
 Subject: [PATCH 2/5] snappy-apparmor-tweaks
 
+Signed-off-by: Lincoln Wallace <lincoln.wallace@canonical.com>
 ---
- profiles/apparmor/apparmor.go | 18 ++++++++++++++++++
+ profiles/apparmor/apparmor.go | 17 +++++++++++++++++
  profiles/apparmor/template.go |  8 ++++++++
- 2 files changed, 26 insertions(+)
+ 2 files changed, 25 insertions(+)
 
 diff --git a/profiles/apparmor/apparmor.go b/profiles/apparmor/apparmor.go
-index d0f2361605..5a0a371d9b 100644
+index 277c853ebe..92e3872722 100644
 --- a/profiles/apparmor/apparmor.go
 +++ b/profiles/apparmor/apparmor.go
-@@ -7,7 +7,9 @@ import (
- 	"bufio"
- 	"io"
+@@ -9,6 +9,7 @@ import (
  	"os"
-+	"os/exec"
+ 	"os/exec"
  	"path"
-+	"strconv"
++  "strconv"
  	"strings"
  	"text/template"
- 
-@@ -27,6 +29,8 @@ type profileData struct {
+ )
+@@ -26,6 +27,8 @@ type profileData struct {
  	Imports []string
  	// InnerImports defines the apparmor functions to import in the profile.
  	InnerImports []string
@@ -31,7 +30,7 @@ index d0f2361605..5a0a371d9b 100644
  }
  
  // generateDefault creates an apparmor profile from ProfileData.
-@@ -46,6 +50,20 @@ func (p *profileData) generateDefault(out io.Writer) error {
+@@ -45,6 +48,20 @@ func (p *profileData) generateDefault(out io.Writer) error {
  		p.InnerImports = append(p.InnerImports, "#include <abstractions/base>")
  	}
  
@@ -53,10 +52,10 @@ index d0f2361605..5a0a371d9b 100644
  }
  
 diff --git a/profiles/apparmor/template.go b/profiles/apparmor/template.go
-index 5dcf35bf45..c9bae7187d 100644
+index 8dbc1b6102..c908b98999 100644
 --- a/profiles/apparmor/template.go
 +++ b/profiles/apparmor/template.go
-@@ -51,5 +51,13 @@ profile {{.Name}} flags=(attach_disconnected,mediate_deleted) {
+@@ -55,5 +55,13 @@ profile {{.Name}} flags=(attach_disconnected,mediate_deleted) {
  
    # suppress ptrace denials when using 'docker ps' or using 'ps' inside a container
    ptrace (trace,read,tracedby,readby) peer={{.Name}},
@@ -71,5 +70,5 @@ index 5dcf35bf45..c9bae7187d 100644
  }
  `
 -- 
-2.25.1
+2.43.0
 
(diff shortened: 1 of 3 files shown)


patches/engine/0001-snappy-apparmor-tweaks.patch
farshidtz, line 1: Please keep the original name and give an equivalent subject to the patch i.e. `[PATCH 2/4] ...`

patches/engine/0001-snappy-buildkit-git-environ.patch
farshidtz, line 1: Rename to `0003-snappy-buildkit-git-environ.patch`. Set an equivalent subject.

patches/engine/0001-snappy-buildkit-git-environ.patch
```diff
@@ -0,0 +1,27 @@
...
+index 5c35f9365b..b0e0fb3aef 100644
+--- a/vendor/github.com/moby/buildkit/util/gitutil/git_cli.go
++++ b/vendor/github.com/moby/buildkit/util/gitutil/git_cli.go
+@@ -183,6 +183,9 @@ func (cli *GitCLI) Run(ctx context.Context, args ...string) (_ []byte, err error
+ 
+ 		cmd.Env = []string{
+ 			"PATH=" + os.Getenv("PATH"),
++      "LD_LIBRARY_PATH=" + os.Getenv("LD_LIBRARY_PATH"),
++      "GIT_EXEC_PATH=" + os.Getenv("GIT_EXEC_PATH"),
++      "GIT_TEMPLATE_DIR=" + os.Getenv("GIT_TEMPLATE_DIR"),
```
farshidtz, line 21: Inconsistent indentation.

snap/snapcraft.yaml
```diff
@@ -232,8 +232,8 @@ parts:
   runc:
     plugin: make
     source: https://github.com/opencontainers/runc.git
-    # from https://github.com/moby/moby/releases/tag/v24.0.5
-    source-tag: v1.1.12
+    # from https://github.com/moby/moby/releases/tag/v27.0.0-rc.2
```
farshidtz, line 235: Basing this on the release notes of a RC is questionable. How about linking to https://github.com/moby/moby/blob/v27.2.0/Dockerfile#L290 ?

snap/snapcraft.yaml
```diff
@@ -336,8 +336,8 @@ parts:
   buildx:
     plugin: nil
     source: https://github.com/docker/buildx.git
-    # https://github.com/moby/moby/releases/tag/v24.0.5
-    source-tag: v0.11.2
+    # https://github.com/moby/moby/releases/tag/v27.1.0
```
farshidtz, line 339: Mismatching version causes confusion. I'd use https://github.com/moby/moby/blob/v27.2.0/Dockerfile#L15

snap/snapcraft.yaml
```diff
@@ -191,8 +191,8 @@ parts:
       install -T bundles/dynbinary-daemon/dockerd "$CRAFT_PART_INSTALL/bin/dockerd"
       # install docker-proxy previously provided by libnetwork part
       install -T bundles/dynbinary-daemon/docker-proxy "$CRAFT_PART_INSTALL/bin/docker-proxy"
-    # https://github.com/moby/moby/blob/v24.0.5/Dockerfile.simple (Docker-supported Go version for Engine)
-    build-snaps: &go ['go/1.20/stable']
+    # https://github.com/moby/moby/blob/v27.2.0/Dockerfile.simple (Docker-supported Go version for Engine)
```
farshidtz, line 194: Should this be based on Dockerfile.simple or Dockerfile? 
In any case, better link to the right line: https://github.com/moby/moby/blob/v27.2.0/Dockerfile#L3

snap/snapcraft.yaml
```diff
@@ -215,8 +215,8 @@ parts:
   containerd:
     plugin: make
     source: https://github.com/containerd/containerd.git
-    # from https://github.com/moby/moby/releases/tag/v24.0.5
-    source-tag: v1.6.21
+    # from https://github.com/moby/moby/releases/tag/v27.2.0
```
farshidtz, line 218: For consistency if you apply other comments: https://github.com/moby/moby/blob/v27.2.0/Dockerfile#L199

snap/snapcraft.yaml
```diff
@@ -295,7 +295,7 @@ parts:
     plugin: cmake
     source: https://github.com/krallin/tini.git
     source-type: git
-    # from https://github.com/moby/moby/blob/v24.0.5/hack/dockerfile/install/tini.installer
+    # from https://github.com/moby/moby/blob/v27.2.0/hack/dockerfile/install/tini.installer
```
farshidtz, line 298: Maybe also add line number: https://github.com/moby/moby/blob/v27.2.0/hack/dockerfile/install/tini.installer#L6

farshidtz (changes requested):
Thanks a lot. Great to have everything upgraded.

I've added some comments about the patch structuring and comments.

Need to also test the upgraded versions together with the Nvidia toolkit.

patches/engine/0001-snappy-buildkit-git-environ.patch
locnnil: This is tricky! I tried a few times to get this indentation right, but once the patch is formatted/generated it didn't get right. And manually edditing the patch seems wrong/hacking for me.

But since this is GO and not Python, I don't see a problem here.

snap/snapcraft.yaml
locnnil: It's not based on an RC, I'm saying that: 

In the RC the version was changed. But the version still the same for the non-rc version. I'm just pointing the place where the version for the Opencontainers get bumped, only for reference. It gets applied for all latter non-rc versions as well. 

The difference is that, once the version is bumped on the e.g `v1.1` it shows up on the release notes, but it didn't show again in the e.g `v1.2`

snap/snapcraft.yaml
locnnil: Same explanation as here: https://github.com/canonical/docker-snap/pull/178#discussion_r1740964988

snap/snapcraft.yaml
farshidtz: I decision made for an RC may get reverted before the actual release, so this isn't a good reference. 

> The difference is that, once the version is bumped on the e.g v1.1 it shows up on the release notes, but it didn't show again in the e.g v1.2

That's why I don't think it is a good idea to chase release notes while there is exists a canonical reference for all these versions: the Dockerfile.

patches/engine/0001-snappy-buildkit-git-environ.patch
farshidtz: So Go code doesn't need to be indented because the compiler doesn't care? :)

lucaskanashiro (approved):
Thanks for this PR. The changes LGTM and the snap is still building fine. As @farshidtz said, we just need to do some testing before landing this.

locnnil:
The successful run: 

https://github.com/canonical/docker-snap/actions/runs/10674609671

Use the artifact from:  https://github.com/canonical/docker-snap/actions/runs/10671266078 

Which means: The Nvidia tests passed for this PR.

patches/engine/0002-snappy-apparmor-tweaks.patch
```diff
@@ -1,28 +1,27 @@
...
-@@ -7,7 +7,9 @@ import (
- 	"bufio"
- 	"io"
+@@ -9,6 +9,7 @@ import (
  	"os"
-+	"os/exec"
+ 	"os/exec"
  	"path"
-+	"strconv"
++  "strconv"
```
farshidtz, line 20:
```suggestion
+	"strconv"
```

farshidtz:
The snap is available for testing on the `latest/edge/runid-10671266078` channel.

To upgrade:
```
sudo snap refresh docker --channel=latest/edge/runid-10671266078
```

farshidtz (approved):
Thanks, we've been testing this in various settings, on arm64 and amd64, also with compose. The upgrade seems to work fine and there are no regressions.

Please squash and merge.

PR #199, opened 2024-12-14
add: utility script to upgrade docker-snap version

locnnil:
- It works by running it from the base root of the project:
```
./upgrade.sh
```
It generates the necessary changes on `snap/snapcraft.yaml` by fetching the latest version available for Moby
using the GitHub API.
And then, based on the Dockerfile for the latest version, it fetches the necessary tags for the other parts of docker-snap. 
It also updates the comments with the new version for the Moby Engine.

--- 

**PS:** It has well known that the script generates some undesired changes like: **It deletes the blank lines on the `snap/snapcraft.yaml`** file, this is because of how `yq` behaves (since it serializes and deserializes the `YAML` file). And it's not something that I'm willing to fix on this script with this PR. The undesired changed can be removed by doing `git restore snap/snapcraft.yaml -p` and selecting the changes that you want to discard. Or also by doing `git add -p` and adding to the stage area only the changes that you want and discarding the other ones with `git restore .`.


.github/workflows/upgrade.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
+
+set -eux
+
+fetch_latest(){
+  # Fetch latest version from Github API
+  #  NOTE: It uses the index number 2 instead of 0 because on moby
+  #  there are the tags xdocs-v1.10-28-mar-2016 and xdocs-v1.10-09-05-2016
+  #  being listed first
+ LATEST=$(curl -s "https://api.github.com/repos/moby/moby/tags" | jq -r '.[2].name')
```
farshidtz, line 10: This is very fragile. How about getting the latest tag via git or using the [releases API](https://docs.github.com/en/rest/releases/releases)?

farshidtz (changes requested):
Nice utility script. Not sure if this should under workflows, considering a workflow for this doesn't exist yet.

upgrade.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+
+set -eux
+
+fetch_latest() {
+  # Fetch latest version from Github releases API
+  LATEST=$(curl -s "https://api.github.com/repos/moby/moby/releases" | jq -r '.[0].name')
```
farshidtz, line 7: `.[0].name` is the release name. The rest of the logic expects the tag name.


```suggestion
  LATEST=$(curl -s "https://api.github.com/repos/moby/moby/releases" | jq -r '.[0].tag_name')
```

farshidtz (changes requested):

upgrade.sh
farshidtz: Also, since it needs just one release, limit the requested numbers.

locnnil: Done here: https://github.com/canonical/docker-snap/pull/199/commits/c54813ebd3de10f9dfcf8a65f1fec05d85279564

farshidtz (approved):
Looks good. Thanks.
I noticed that this is not upgrading the Nvidia runtime parts. Would be useful to also upgrade nvidia-container-toolkit and libnvidia-container, but that needs to rely on a reference (major version number?) to avoid breaking changes.

PR #200, opened 2024-12-14
ci: Release on edge the latest version every two weeks

locnnil:
If the smoke tests passes, it releases to `beta`.

- Depends on: https://github.com/canonical/docker-snap/pull/199

### DIFF
--- a/.github/workflows/upgrade.sh
+++ b/.github/workflows/upgrade.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+
+set -eux
+
+fetch_latest(){
+  # Fetch latest version from Github API
+  #  NOTE: It uses the index number 2 instead of 0 because on moby
+  #  there are the tags xdocs-v1.10-28-mar-2016 and xdocs-v1.10-09-05-2016
+  #  being listed first
+ LATEST=$(curl -s "https://api.github.com/repos/moby/moby/tags" | jq -r '.[2].name')
+}
+
+# Validate the version format
+validate_version(){
+  # Original simplified RegEx:
+  # v\d+.\d+.\d+\-*(rc.\d|rc\d|beta.\d)*
+  # By analysing the last tags on github.com/moby/moby/tags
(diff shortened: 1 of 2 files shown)


.github/workflows/upgrade.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
+
+set -eux
+
+fetch_latest(){
+  # Fetch latest version from Github API
+  #  NOTE: It uses the index number 2 instead of 0 because on moby
+  #  there are the tags xdocs-v1.10-28-mar-2016 and xdocs-v1.10-09-05-2016
+  #  being listed first
+ LATEST=$(curl -s "https://api.github.com/repos/moby/moby/tags" | jq -r '.[2].name')
```
farshidtz, line 10: This will be very fragile. What if a third such tag gets added?
I suggest relying on the most recent tag using git or [release](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28) instead.

farshidtz (changes requested):

farshidtz:
Thanks for the initiative. 
I support the idea of automating for the upgrade, but the proposed method is a bit problematic as it builds and releases the snap with code that is not checked in, making it hard to audit the artifact against sources. With this release workflow, the changed code never even makes it back to the upstream.

Instead, I propose a semi-automated approach with automated code changes submitted via a PR. This is a well established pattern used by automated dependency update tools such as Dependabot and Renovate. The PR should allow maintainers to perform QA on the changes and the build. Once approved and merged, the remaining automation can take place for the release.

locnnil:
> Instead, I propose a semi-automated approach with automated code changes submitted via a PR. This is a well established pattern used by automated dependency update tools such as Dependabot and Renovate. The PR should allow maintainers to perform QA on the changes and the build. Once approved and merged, the remaining automation can take place for the release.

Sure, I also like that approach.

farshidtz:
Thanks. I'll close this then.

PR #206, opened 2025-01-06
fix: logic inconsistency for `publish` input on nvidia-test workflow

locnnil:
The use of `$SNAP_CHANNEL` as `latest/edge/runid-<RUN_ID>` only makes sense if the checkbox to `publish` is marked as true.

If not, it means that it needs to run against `latest/edge` but there is no logic handling it and the workflow fails with the error:
```
 error: requested a non-existing branch on latest/edge for snap "docker":
```
as shown here:
https://github.com/canonical/docker-snap/actions/runs/12631899123/job/35194514402#step:5:1469

### DIFF
--- a/.github/workflows/nvidia-test.yml
+++ b/.github/workflows/nvidia-test.yml
@@ -42,7 +42,7 @@ jobs:
         env:
           TESTFLINGER_DIR: .github/workflows/testflinger
           JOB_QUEUE: docker-nvidia
-          SNAP_CHANNEL: latest/edge/runid-${{ inputs.run_id }}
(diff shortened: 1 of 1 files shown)


farshidtz:
Thanks for the PR.

> If not, it means that it needs to run against `latest/edge`

The use case for unsetting publish is not to run it against `latest/edge`. The checkbox was added (https://github.com/canonical/docker-snap/commit/bd0f133d84fff2cc4722ed592939d3746f8ad96e) to allow re-testing a pre-build artifact from a previous run, for example when developing the tests or the CI. Without this, the job will try to re-upload the same artifact and [fail](https://github.com/canonical/docker-snap/actions/runs/12369291180/job/34520883816) due to the Store's uniqueness check.

We could add support for testing a build from an arbitrary channel, not just `edge/runid-${{ inputs.run_id }}` but a couple of things need to change to make that work.

Issue #181, opened 2024-09-06
App Armor denials on Smoke test `Hello World (journald)` on Noble.


locnnil:
The smoke-test: 
https://github.com/canonical/docker-snap/blob/a41145de895afc7602cf994ebee855e1c63e7be8/.github/workflows/smoke-test.yml#L112-L119

Fails when running on Noble (but not on Focal and Jammy), as shown here:

Docker Snap **v24.0.5**:
https://github.com/canonical/docker-snap/actions/runs/10738846000

```log
[  210.260771] eth0: renamed from vethfac58ea
[  210.277728] docker0: port 1(vethf07816c) entered blocking state
[  210.277734] docker0: port 1(vethf07816c) entered forwarding state
[  210.333867] vethfac58ea: renamed from eth0
[  210.344540] docker0: port 1(vethf07816c) entered disabled state
[  210.35[222](https://github.com/canonical/docker-snap/actions/runs/10738846000/job/29783725594#step:10:223)0] docker0: port 1(vethf07816c) entered disabled state
[  210.352923] vethf07816c (unregistering): left allmulticast mode
[  210.352928] vethf07816c (unregistering): left promiscuous mode
[  210.352932] docker0: port 1(vethf07816c) entered disabled state
Sep 06 13:10:55 fv-az1693-731 kernel: audit: type=1400 audit(1725628255.588:263): apparmor="DENIED" operation="ptrace" class="ptrace" profile="snap.docker.dockerd" pid=5269 comm="ps" requested_mask="read" denied_mask="read" peer="/usr/sbin/haveged"
Sep 06 13:10:55 fv-az1693-731 kernel: audit: type=1400 audit(1725628255.590:264): apparmor="DENIED" operation="ptrace" class="ptrace" profile="snap.docker.dockerd" pid=5269 comm="ps" requested_mask="read" denied_mask="read" peer="rsyslogd"
Sep 06 13:10:55 fv-az1693-731 kernel: audit: type=1400 audit(1725628255.591:265): apparmor="DENIED" operation="ptrace" class="ptrace" profile="snap.docker.dockerd" pid=5269 comm="ps" requested_mask="read" denied_mask="read" peer="/usr/sbin/chronyd"
[...]

PR #207, opened 2025-01-06
ci: add Ubuntu 20.04 and 24.04 to test runners

locnnil:
- Related to https://github.com/canonical/docker-snap/issues/181

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -89,6 +89,7 @@ jobs:
           sudo snap connect docker:network-control :network-control
           sudo snap connect docker:support :docker-support
           sudo snap connect docker:privileged :docker-support
+          sudo snap connect docker:log-observe
 
           sudo snap stop --disable docker.dockerd # for good measure
           sleep 1
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -56,9 +56,6 @@ jobs:
 
       - name: refresh snapd
         run: |
-          # It seems that installing the deb version of snapd wasn't enought
-          # to have the snapd snap present (or at least active)
-          # sudo snap install hello-world
           sudo snap install snapd
           sudo snap refresh snapd
           sudo snap list
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -33,7 +33,7 @@ jobs:
 
   smoke-test:
     needs: build
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -54,12 +54,27 @@ jobs:
...
 
-          sudo apt update
-          sudo apt install -y snapd
-          sudo snap refresh
-          sudo snap refresh --channel beta snapd
+      - name: refresh snapd
+        run: |
+          sudo snap install snapd
+          sudo snap refresh snapd
           sudo snap list
```
farshidtz, line 61: Refresh is unnecessary since `snap install` has just installed the latest/stable.
```suggestion
      # The image comes with a slightly outdated snapd Debian package
      - name: Install latest snapd as a snap
        run: |
          sudo snap install snapd
          sudo snap list
```

.github/workflows/smoke-test.yml
```diff
@@ -54,12 +54,27 @@ jobs:
...
+          # System and snapd data
+          snap version
+
+          # Check how snapd is installed
+          sudo snap debug execution snap
+
+          # Check apparmor abi version
+          sudo snap debug execution apparmor
+
+          # Check avaible sandbox-features
```
farshidtz, line 75:
```suggestion
          # Check available sandbox-features
```

farshidtz (changes requested):

.github/workflows/smoke-test.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
   smoke-test:
     needs: build
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
farshidtz, line 36: How about adding 20.04 and 24.04? The snap should be tested on all.
```suggestion
strategy:
    matrix:
        os: [ubuntu-24.04, ubuntu-22.04, ubuntu-20.04]
    runs-on: ${{ matrix.os }}
```

.github/workflows/smoke-test.yml
```diff
@@ -54,12 +58,27 @@ jobs:
...
+          snap version || true
+
+          # Check how snapd is installed
+          sudo snap debug execution snap || true
+
+          # Check apparmor abi version
+          sudo snap debug execution apparmor || true
+
+          # Check available sandbox-features
+          sudo snap debug sandbox-features || true
```
locnnil, line 80: The version of snapd on Focal don't have the `execution` option on `debug` command, and because this is just information being fetching, we don't want any of those to be a cause of workflow failure.

farshidtz: Can this be done by setting `set +e`?

See also:
- https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrunshell

.github/workflows/smoke-test.yml
```diff
@@ -109,14 +133,16 @@ jobs:
...
+      # NOTE: deactivating test for now since journald is broken for docker-snap on Noble
+        # see: https://github.com/canonical/docker-snap/issues/181
+      # - name: Hello World (journald)
+      #   run: |
+      #     trap 'echo "error, sad day ($?)"; sleep 1; sudo snap logs -n=20 docker.dockerd; sleep 1; sudo tail -n20 /var/log/*.log; sudo dmesg | tail -n20; sudo journalctl --no-pager | grep DENIED | grep docker' ERR
+      #     expectedOutput="testing-journald-log-driver-$RANDOM-$RANDOM"
+      #     sudo docker run --name test-journald --log-driver journald bash -c 'echo "$@"' -- "$expectedOutput"
+      #     actualOutput="$(sudo docker logs test-journald)"
+      #     [ "$actualOutput" = "$expectedOutput" ]
+      #     docker rm test-journald
```
farshidtz, line 145: Going back to this with the added OS matrix: this step can now run conditionally and skip only when matrix.os is ubuntu-24.04.

farshidtz (approved):
LGTM. Thanks.

PR #208, opened 2025-01-07
CI: Test for syslog

locnnil:
Add test for syslog driver.

- Depends on: https://github.com/canonical/docker-snap/pull/207

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -133,6 +133,14 @@ jobs:
       - name: Hello World
         run: sudo docker run --rm hello-world
 
+      - name: Hello World (syslog)
+        run: |
+          trap 'echo -e "\nerror, sad day ($?)"; sleep 1; sudo snap logs -n=20 docker.dockerd; sleep 1; sudo tail -n20 /var/log/*.log; sudo dmesg | tail -n20; sudo journalctl --no-pager | grep DENIED | grep docker' ERR
+          expectedOutput="testing-syslog-log-driver-$RANDOM-$RANDOM"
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -133,18 +134,17 @@ jobs:
...
           trap 'echo "error, sad day ($?)"; sleep 1; sudo snap logs -n=20 docker.dockerd; sleep 1; sudo tail -n20 /var/log/*.log; sudo dmesg | tail -n20; sudo journalctl --no-pager | grep DENIED | grep docker' ERR
-          expectedOutput="testing-journald-log-driver-$RANDOM-$RANDOM"
-          sudo docker run --name test-journald --log-driver journald bash -c 'echo "$@"' -- "$expectedOutput"
-          actualOutput="$(sudo docker logs test-journald)"
+          expectedOutput="testing-log-drivers-$RANDOM-$RANDOM"
+          sudo docker run --name test-log-drivers --log-driver ${{ matrix.log-driver }} bash -c 'echo "$@"' -- "$expectedOutput"
+          actualOutput="$(sudo docker logs test-log-drivers)"
           [ "$actualOutput" = "$expectedOutput" ]
-          docker rm test-journald
+          docker rm test-log-drivers
```
locnnil, line 147: @farshidtz Should I just replicate the test instead of use matrix? I have a feeling that the workflow will look more readable in the end. Because instead of having 6 results for the smoke-tests, we will have only three.

farshidtz: Yes, this is just a separate test unit (job step) but the matrix applies to the whole job.

farshidtz (approved):
Looks good. Thanks.

PR #211, opened 2025-01-08
snap: add missing meta-data

xypron:
Add 'issues', 'source-code', and 'website' tags.
This allows to display the information with `snap info docker`.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -2,6 +2,9 @@ name: docker
 #title: Docker
 version: '27.4.1'
 summary: Docker container runtime
+issues: https://github.com/canonical/docker-snap/issues
+source-code: https://github.com/canonical/docker-snap
+website: https://github.com/canonical/docker-snap
 description: |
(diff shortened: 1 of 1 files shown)


locnnil (approved):
LGTM.

xypron:
I signed the commit now as the CI wanted a verified commit.

farshidtz:
Thanks @xypron but these are not going to have any effect. The metadata of this snap is managed on the Snap Store. I've just added the source URL which was missing there.

We should remove the redundant metadata from the source to avoid confusion.

xypron:
@farshidtz
`snap info` can show contact information but not issues (which looks like a missing feature in snapd). Maybe in the snapstore you could additionally maintain that field with value https://github.com/canonical/docker-snap/issues.

xypron:
Snapd bug filed:

Missing field 'Issues: ' in snap info output (https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/2093419)

PR #213, opened 2025-01-10
docs: Remove snap description

farshidtz:
The snap's metadata is maintained on the Snap Store.

Removing the description in source to ease maintenance of the instructions and avoid confusion (#211).

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,39 +1,13 @@
 name: docker
-#title: Docker
 version: '27.4.1'
 summary: Docker container runtime
-description: |
-  Build and run container images with Docker.
-
-  **Usage**
-
-  * This build can only access files in the home directory. So Dockerfiles and all other files used in commands like `docker build`, `docker save` and `docker load` need to be in $HOME.
-  * You can change the configuration of this build by modifying the files in `/var/snap/docker/current/`.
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -1,39 +1,13 @@
...
-      sudo snap enable docker
-
-  **Warning:** if you add your user to the `docker` group, it will have similar power as the `root` user. For details on how this impacts security in your system, see https://docs.docker.com/engine/security/#docker-daemon-attack-surface
-
-  **Authors**
-
-  This snap is built by Canonical based on source code published by Docker, Inc. It is not endorsed or published by Docker, Inc.
-
-  Docker and the Docker logo are trademarks or registered trademarks of Docker, Inc. in the United States and/or other countries. Docker, Inc. and other parties may also have trademark rights in other terms used herein.
+description: Refer to https://snapcraft.io/docker
```
locnnil, line 4: I think that the first sentence is a good one to leave it in the description. 

```suggestion
description: | 
  Build and run container images with Docker.
  Refer to https://snapcraft.io/docker
```

locnnil (approved):
LGTM, I only have one comment.

snap/snapcraft.yaml
farshidtz: Why is this information needed? 
A user installing docker should already know what to expect. When not, they can see the Summary (Docker container runtime) locally and more details by referring to the store.

PR #215, opened 2025-01-19
CI: Run tests on arm runners for Noble and Jammy

locnnil:


.github/workflows/smoke-test.yml
```diff
@@ -14,7 +14,15 @@ defaults:
...
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        runs:
+          - arch: amd64
+            os: ubuntu-latest
+          - arch: arm64
+            os: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runs.os }}
```
farshidtz, line 25:
```suggestion
      matrix:
        runner:
          - arch: amd64
            os: ubuntu-latest
          - arch: arm64
            os: ubuntu-24.04-arm
    runs-on: ${{ matrix.runner.os }}
```
Resulting in more meaningful input names:
- `runner.os`
- `runner.arch`

To be changed below as well.

locnnil: Done here: https://github.com/canonical/docker-snap/pull/215/commits/fddb35d9e7a1051a9793f18064483a13b7afeb92 thank you!

PR #216, opened 2025-01-20
CI: Run Nvidia workflow on UC22

locnnil:
Currently, the OS used on the Nvidia tests [is hardcoded](https://github.com/canonical/docker-snap/blob/7d5e826d6b6787cfa98ba9d803cdbf319003c942/.github/workflows/testflinger/nvidia-job.yaml#L7):

https://github.com/canonical/docker-snap/blob/7d5e826d6b6787cfa98ba9d803cdbf319003c942/.github/workflows/testflinger/nvidia-job.yaml#L7

However, we already have mechanisms in place to support testing on UC22 as well:

https://github.com/canonical/docker-snap/blob/7d5e826d6b6787cfa98ba9d803cdbf319003c942/.github/workflows/testflinger/scripts/setup.sh#L49-L63

This PR updates the workflow to include tests for UC22 alongside the existing tests for Noble by using a matrix configuration.

### DIFF
--- a/.github/workflows/nvidia-test.yml
+++ b/.github/workflows/nvidia-test.yml
@@ -11,6 +11,13 @@ on:
         description: 'Run id number'
         required: true
         type: number
+      os_release:
+        description: 'Os release'
+        required: true
+        type: choice
+        options:
+          - core22-latest
+          - noble
       publish:
         description: 'Publish to Store'
         default: true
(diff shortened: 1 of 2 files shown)


farshidtz:
@locnnil Please submit from a local branch to allow testing.

IMO, using a build matrix is not the best approach here because it will require multiple GH Runners to work in parallel. We only have one, which means they will be re-deployed for every run, taking several minutes. This might be okay; I'm curious to know how long it takes between the runs.

A slightly modified approach would be to prepare multiple jobs and submit them in a sequence in different steps:
https://github.com/canonical/docker-snap/blob/7d5e826d6b6787cfa98ba9d803cdbf319003c942/.github/workflows/nvidia-test.yml#L63-L67

locnnil:
I'm closing this in favor of https://github.com/canonical/docker-snap/pull/217 submitted from a local branch.

PR #221, opened 2025-01-27
build(upgrade.sh): RegEx to match only commented URL pointing to moby Dockerfile

locnnil:
This part of upgrade script:

https://github.com/canonical/docker-snap/blob/69554a091f95da4015a0f38dbb05d7dd937b151f/upgrade.sh#L113

It is designed to change the comments with URLs pointing to Dockerfile of Moby.

But it only looks for the tag version, that's why it was changing the version of other parts that matches the exact same version.

This change adds more context to the RegEx so it only matches the URLs on the comments.

[skip ci]


farshidtz (approved):
LGTM, thanks.

PR #223, opened 2025-01-28
test: Improve Testflinger Nvidia testing

farshidtz:
- Upgrade CUDA toolkit
- Use build matrix for distros
- Reduce setup logging verbosity
- Improve test logging
- Add snap channel to annotations

### DIFF
--- a/.github/workflows/testflinger/scripts/test.sh
+++ b/.github/workflows/testflinger/scripts/test.sh
@@ -2,8 +2,10 @@
 set -e
 
 # Test nvidia-smi
-smi_test() {
+smi_test() (
   . /etc/os-release
+
+  set -x
   
   if [[ $ID == "ubuntu" ]]; then
     sudo docker run --rm --runtime=nvidia --gpus all --env PATH="${PATH}:/var/lib/snapd/hostfs/usr/bin" ubuntu nvidia-smi || true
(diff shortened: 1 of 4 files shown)


farshidtz:
Here is an example run: https://github.com/canonical/docker-snap/actions/runs/13388583832

The output of `apt` commands are still very verbose, resulting in a few thousand lines of logs.

.github/workflows/testflinger/scripts/test.sh
```diff
@@ -1,14 +1,14 @@
...
 smi_test() {
   . /etc/os-release
-
+  
   if [[ $ID == "ubuntu" ]]; then
-    sudo docker run --rm --runtime=nvidia --gpus all --env PATH="${PATH}:/var/lib/snapd/hostfs/usr/bin" ubuntu nvidia-smi || exit 1
+    sudo docker run --rm --runtime=nvidia --gpus all --env PATH="${PATH}:/var/lib/snapd/hostfs/usr/bin" ubuntu nvidia-smi || true
   elif [[ $ID == "ubuntu-core" ]]; then
-    sudo docker run --rm --runtime nvidia --gpus all ubuntu bash -c "/snap/docker/*/graphics/bin/nvidia-smi" || exit 1
+    sudo docker run --rm --runtime nvidia --gpus all ubuntu bash -c "/snap/docker/*/graphics/bin/nvidia-smi" || true
```
locnnil, line 11: I don't understand the reason of `|| true` wouldn't this make these tests always pass? Do we want that?

locnnil (commented):
Looks great, I only have a concern regarding the `smi_test`.

.github/workflows/testflinger/scripts/test.sh
farshidtz: This step is useful for debugging, based on the nvidia-smi's output, but it doesn't run any GPU workload. Letting it pass regardless of the exit code allows running the next test. If the nvidia runtime fails, the next test will catch it.

I agree that it is not ideal. It would be better to implement the logic to let all tests run but exit with non-zero code if either of them fail.

locnnil (approved):
LGTM +1!

locnnil (dismissed):
Thanks for the changing to use `apt-get`. LGTM +1.

farshidtz:
Example run with the latest changes: https://github.com/canonical/docker-snap/actions/runs/13446514132/job/37572947675

locnnil (approved):
LGTM +1!

I would suggest redirecting the std output of `apt-get install` commands as well, so it only shows up the std error with the errors and warnings, but I leave it up to you.

Issue #183, opened 2024-09-20
Build warning regarding `opengl` and `removable-media` plugs


farshidtz:
There is a warning in the snapcraft logs right after the snap creation:
```
Warning: implicit plug assignment in 'opengl' and 'removable-media'. Plugs should be assigned to the app to which they apply, and not implicitly assigned via the global 'plugs:' stanza which is intended for configuration only.
(Reference: https://snapcraft.io/docs/snapcraft-top-level-metadata#heading--plugs-and-slots-for-an-entire-snap)
```

The `opengl` plug is only defined globally, but `removable-media` set also as plug for the `docker` app.

PR #224, opened 2025-01-29
build: Set removable-media and opengl plugs only where is needed

locnnil:
Fix  Snapcraft build warning:
```
Warning: implicit plug assignment in 'opengl' and 'removable-media'.
Plugs should be assigned to the app to which they apply, and not implicitly
assigned via the global 'plugs:' stanza which is intended for configuration only.
```

- Fixes: https://github.com/canonical/docker-snap/issues/183

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -97,12 +97,14 @@ apps:
       - docker-cli
       - network
       - home
+      - removable-media
 
   nvidia-container-toolkit:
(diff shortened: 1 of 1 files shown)


farshidtz:
This has been reported here: https://github.com/canonical/docker-snap/issues/183

farshidtz (approved):

locnnil:
With this change, it has the same effect as defining it globally but eliminates the warning.

PR #230, opened 2025-02-04
Split publish

locnnil:
This PR aims to:

* Publish based on artifact built from a PR
* Changed the Nvidia Tests to receive the channel to run the tests


locnnil:
Not sure why, but I cannot see the `publish.yml` as a workflow option. 

image

farshidtz:
> Not sure why, but I cannot see the `publish.yml` as a workflow option.
> 
> image

This is because the workflow is not on the default (main) branch.

PR #238, opened 2025-02-20
Upgrade to 28.0.0

locnnil:


locnnil:
The error seems to be on applying this patch: 

https://github.com/canonical/docker-snap/blob/fa9169f8c9715f8c3faadf97e8b412a0336a81c0/patches/engine/0004-snappy-containerd-sock.patch#L6-L23

The breaking change on containerd was introduced this commit: https://github.com/containerd/containerd/commit/f6e30e9622b79c1e3ef64e22329bbabe6d1789e7

PR #252, opened 2025-03-06
doc: docker aliases 

locnnil:
Add Documentation regarding the existence of Docker alias `docker-compose` poiting to the `compose` app.

[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ sudo snap connect docker:home
 ```
 
 The `docker-compose` [alias](https://snapcraft.io/docs/commands-and-aliases) was set automatically for Compose V1 and remains for backwards-compatiblity.
-The recommended command-line syntax since Compose V2 is `docker compose` as described on the [Docker's official documentation](https://docs.docker.com/compose/releases/migrate/).
+The recommended command-line syntax since Compose V2 is `docker compose` as described [here](https://docs.docker.com/compose/releases/migrate/).
 
 ### Changing the data root directory
 
--- a/README.md
+++ b/README.md
@@ -33,6 +33,9 @@ If you are using Ubuntu Core 16, connect the `docker:home` plug as it's not auto
 sudo snap connect docker:home
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,6 +61,21 @@ Then restart the dockerd service:
...
+
+You can verify this by running:  
+
+```console
+$ snap aliases docker
+Command         Alias           Notes
+docker.compose  docker-compose  -
+```  
+
+Although the [migration to Compose V2](https://docs.docker.com/compose/releases/migrate/) deprecates the `docker-compose` command in favor of `docker compose`, the alias still exists for backward compatibility.
```
farshidtz, line 77: I'd simplify and move this to same section with connections. Example:
> The `docker-compose` [alias](https://snapcraft.io/docs/commands-and-aliases) was set automatically for Compose V1 and remains for backwards-compatiblity. The recommended command-line syntax since Compose V2 is `docker compose` [[ref](https://docs.docker.com/compose/releases/migrate/)].

farshidtz (changes requested):
Thanks.
I suggest moving the instruction.

We should also add info to the snapcraft file to make it clear that the app has this alias.

README.md
```diff
@@ -33,6 +33,9 @@ If you are using Ubuntu Core 16, connect the `docker:home` plug as it's not auto
 sudo snap connect docker:home
 ```
 
+The `docker-compose` [alias](https://snapcraft.io/docs/commands-and-aliases) was set automatically for Compose V1 and remains for backwards-compatiblity.
+The recommended command-line syntax since Compose V2 is `docker compose` as described on the [Docker's official documentation](https://docs.docker.com/compose/releases/migrate/).
```
farshidtz, line 37: Same link title given to two different URLs. On line 8, "Docker's official documentation" links to https://docs.docker.com, but here it links to "Migrate to Compose V2". A reader will probably think these are the same references. In general, longer sentences and more text make it more difficult to find the right information.

```suggestion
The recommended command-line syntax since Compose V2 is `docker compose` as described [here](https://docs.docker.com/compose/releases/migrate/).
```

farshidtz (changes requested):

farshidtz (approved):

PR #255, opened 2025-03-07
chore(upgrade.sh): Add function preventing yq from deleting blank lines

locnnil:
`yq` generates unnecessary diff by discarding blank lines during yaml unmarshalling, a side effect of [gopkg.in/yaml.v3](https://pkg.go.dev/gopkg.in/yaml.v3) library.

This adds the function `yq_pb` that:

1. Takes the `yq` expression and the target file as argument and passes to yq;
2. Generates a diff of the yaml file compared to the yq change ignoring the blank lines and blank spaces (`-B` and `-w`);
3. Apply the patch back to the original yaml file.

# Testing:

For testing follow the steps:
```shell
# Fetch the changes from this PR:
git clone git@github.com:canonical/docker-snap.git && cd docker-snap
git fetch origin pull/255/head:fix-yq-on-upgrade
git checkout fix-yq-on-upgrade


# revert and stash the changes of last commit:
git reset --soft HEAD^
git stash

# Go to the commit: d28b3d427 (or any older commit)
git checkout d28b3d427
git stash pop

# Run the script:
./upgrade.sh

# Check the clean diff:
git diff snap/snapcraft.yaml
```

### DIFF
--- a/upgrade.sh
+++ b/upgrade.sh
@@ -2,6 +2,11 @@
 
 set -eux
 
+# yq preserving blank lines
+yq_pb() {
+  yq "$1" "$2" | diff -wB "$2" - | patch "$2" -
+}
+
 fetch_latest() {
   # Fetch latest version from Github releases API
   LATEST=$(curl -s "https://api.github.com/repos/moby/moby/releases?per_page=1" | jq -r '.[0].tag_name')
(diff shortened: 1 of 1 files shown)


upgrade.sh
```diff
@@ -2,6 +2,11 @@
 
 set -eux
 
+# yq preserving blank lines
+yq_pb() {
```
farshidtz, line 6: The acronym is hard to guess, and not very useful. `yq_p` is sufficient to indicate that is a function and not the command.

locnnil: Done here: https://github.com/canonical/docker-snap/pull/255/commits/da07a18356f15aeebfa7bdbeb25651251dc70eb4

farshidtz (approved):
LGTM. Thanks.
This is a [known issue](https://github.com/mikefarah/yq/issues/515), and the implemented solution is inline with other proposals on that issue.

PR #258, opened 2025-03-19
ci(smoke-test): Regression test to check non-stopping container problem

locnnil:
There is no smoke-test for testing the ability to **stop containers**, and **check** if:

- https://github.com/canonical/docker-snap/blob/4418a554b08359e4b38835049f578048ae517e0a/bin/dockerd-wrapper#L4-L14
- https://github.com/canonical/docker-snap/blob/4418a554b08359e4b38835049f578048ae517e0a/patches/engine/0001-snappy-aa-profile-reload.patch#L27

Still being effective.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -182,7 +182,7 @@ jobs:
       # NOTE: regression test for aa default profile reload issue 
       # see: https://github.com/canonical/docker-snap/issues/36
       - name: Stop container
-        timeout-minutes: 3
+        timeout-minutes: 1
         run: |
           trap 'echo "error, sad day ($?)"; sleep 1; sudo snap logs -n=20 docker.dockerd; sleep 1; sudo tail -n20 /var/log/*.log; sudo dmesg | tail -n20; sudo journalctl --no-pager | grep DENIED | grep docker' ERR
           docker run --rm -d --name test bash bash -c 'cleanup(){ sleep 60; }; trap cleanup SIGTERM SIGINT SIGHUP SIGQUIT; while true; do sleep 1; done'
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -179,6 +179,15 @@ jobs:
...
         run: sudo docker run --rm --init hello-world
 
+      # NOTE: regression test for aa default profile reload issue 
+      # see: https://github.com/canonical/docker-snap/issues/36
+      - name: Stop container
+        run: |
+          trap 'echo "error, sad day ($?)"; sleep 1; sudo snap logs -n=20 docker.dockerd; sleep 1; sudo tail -n20 /var/log/*.log; sudo dmesg | tail -n20; sudo journalctl --no-pager | grep DENIED | grep docker' ERR
+          docker run --rm -d --name test bash bash -c 'cleanup(){ sleep 60; }; trap cleanup SIGTERM SIGINT SIGHUP SIGQUIT; while true; do sleep 1; done'
+          sleep 5
+          sudo docker stop test
```
farshidtz, line 189: This step could benefit from a [timeout](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes) to make sure the container ends up getting killed in a reasonable time (i.e. 10s after calling docker stop).

farshidtz (approved):
Thanks for adding the tests.

.github/workflows/smoke-test.yml
locnnil: Added here: https://github.com/canonical/docker-snap/pull/258/commits/5c984b49a5e39b298dc379efa44c42bed989ebc7

According to GH docs, it isn't possible to define a fractional time, it must be specified in **minutes** as **integer** **values**.

farshidtz (approved):

Issue #281, opened 2025-06-02
docker: failed to register layer: mkdir /snap: permission denied


mr-cal:
### Bug description

The docker snap started failing on Ubuntu 22.04 with the error `docker: failed to register layer: mkdir /snap: permission denied`.  

### To reproduce

```bash
> multipass launch jammy --name vm1 --disk 20G
Launched: vm1

> multipass shell jammy

> > sudo snap install docker
docker 28.1.1 from Canonical✓ installed

> > sudo docker run -it ghcr.io/canonical/snapcraft:8_core24@sha256:485d867971b9b310889b0b76bc8329f1ff8087f9c68101cbfad357ee3407100c
Unable to find image 'ghcr.io/canonical/snapcraft:8_core24@sha256:485d867971b9b310889b0b76bc8329f1ff8087f9c68101cbfad357ee3407100c' locally
ghcr.io/canonical/snapcraft@sha256:485d867971b9b310889b0b76bc8329f1ff8087f9c68101cbfad357ee3407100c: Pulling from canonical/snapcraft
eda6120e237e: Pull complete
a8f927883af9: Extracting [==================================================>]  173.9MB/173.9MB
92f95457a9ff: Download complete
9beec1e38464: Download complete
docker: failed to register layer: mkdir /snap: permission denied

> > journalctl --since "4 minutes ago" --no-pager
Jun 02 10:58:05 vm1 systemd[1]: snap.docker.docker-b2585034-ef21-4a13-8258-d7fed046c750.scope: Deactivated successfully.
Jun 02 10:58:18 vm1 sudo[3402]:   ubuntu : TTY=pts/0 ; PWD=/home/ubuntu ; USER=root ; COMMAND=/snap/bin/docker run -it -v /home/ubuntu:/project ghcr.io/canonical/snapcraft:8_core24
Jun 02 10:58:18 vm1 sudo[3402]: pam_unix(sudo:session): session opened for user root(uid=0) by ubuntu(uid=1000)
Jun 02 10:58:18 vm1 systemd[1]: Started snap.docker.docker-86fb233d-d18f-465c-b521-0ebe9f37c977.scope.
Jun 02 10:58:29 vm1 audit[2997]: AVC apparmor="DENIED" operation="mkdir" profile="snap.docker.dockerd" name="/snap/" pid=2997 comm="dockerd" requested_mask="c" denied_mask="c" fsuid=0 ouid=0
[...]

PR #284, opened 2025-06-12
fix: add patch for avoid use of pivot_root() during image layer extraction

locnnil:
- fixes: https://github.com/canonical/docker-snap/issues/281

On the moby code [vendor/github.com/moby/go-archive/internal/mounttree/switchroot_linux.go](https://github.com/moby/moby/blob/bcbcbb73fa7138fa3af113da3e27ac438f73e230/vendor/github.com/moby/go-archive/internal/mounttree/switchroot_linux.go) the func `SwitchRoot()` performs the changing of the root filesystem for a given process. Inside the function currently is used a [`pivot_root(2)`](https://man7.org/linux/man-pages/man2/pivot_root.2.html) syscall (which wasn't the case in the past) and then this operation is being used for image layer extraction.

After the `pivot_root(2)` syscall is made, AppArmor re-intereprets all paths to the new root, (which isn't the case for `chroot(2)`), so aa is interpreting `/var/snap/docker/common/var-lib-docker/overlay2/<hash>/diff/.pivot_root<rand-num>/` as being `/`, then operations like `mkdir /snap` fail because aren't allowed by `snap.docker.dockerd` profile.

 This patch aims to force the usage of the fallback option available in this same code, with the [`chroot(2)`](https://man7.org/linux/man-pages/man2/chroot.2.html) syscall.

## Testing

-  install the docker revision **3262**
```shell
sudo snap install docker --revision=3262
```
- Pull the image `ghcr.io/locnnil/drc:pts`

```shell
sudo docker pull ghcr.io/locnnil/drc:pts
```

### Test results

```console
ubuntu@vmm:~$ sudo snap install docker --revision=3262
docker 28.1.1-fix from Canonical✓ installed
ubuntu@vmm:~$
ubuntu@vmm:~$ snap list docker
Name    Version     Rev   Tracking       Publisher   Notes
docker  28.1.1-fix  3262  latest/stable  canonical✓  -
ubuntu@vmm:~$
ubuntu@vmm:~$ sudo docker pull ghcr.io/locnnil/drc:pts
pts: Pulling from locnnil/drc
ea6b5193b26e: Pull complete
60d1f517012e: Pull complete
Digest: sha256:cc10074b9c2c0f157027234d8ab39f55ebcfc505bf026949395d5df9017549e5
Status: Downloaded newer image for ghcr.io/locnnil/drc:pts
ghcr.io/locnnil/drc:pts
```

Journal logs:
```
Jun 11 23:49:17 vmm sudo[15185]:   ubuntu : TTY=pts/5 ; PWD=/home/ubuntu ; USER=root ; COMMAND=/snap/bin/docker pull ghcr.io/locnnil/drc:pts
Jun 11 23:49:17 vmm sudo[15185]: pam_unix(sudo:session): session opened for user root(uid=0) by ubuntu(uid=1000)
Jun 11 23:49:17 vmm systemd[1]: Started snap.docker.docker-3f342048-5c90-4f50-bc7c-2ca9424496b2.scope.
Jun 11 23:49:19 vmm systemd[1]: snap.docker.docker-3f342048-5c90-4f50-bc7c-2ca9424496b2.scope: Deactivated successfully.
Jun 11 23:49:19 vmm sudo[15185]: pam_unix(sudo:session): session closed for user root
```

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -200,7 +200,7 @@ jobs:
           sudo docker run --rm debian-hello
           sudo docker image rm debian-hello
 
-      # NOTE: regression test for check patch that forces
+        # NOTE: regression test for check patch that forces
         # use of chroot(2) syscall instaed of pivot_root(2)
         # see: https://github.com/canonical/docker-snap/issues/281
       - name: Pull image that write to /snap


locnnil:
> Do we have any idea if we can rely on this being there forever, or will docker eventually remove it and only use pivot_root?

I think that it's very unlikely that they would remove it, unless there is some CVE or some other minor security vulnerability with some proven exploit using the `chroot(2)`, disregarding a case involving a security issue it's very unlikely because this is a fallback which ensures that it still works even if `pivot_root(2)` fails, and it can fail for any reason, like being blocked by seccomp for example (I know that the [docker_support](https://github.com/canonical/snapd/blob/master/interfaces/builtin/docker_support.go) profile [already allows](https://github.com/canonical/snapd/blob/61f2a34ddba1bfb4fdc5dca08d2c665d954fc552/interfaces/builtin/docker_support.go#L731) `pivot_root(2)` I'm talking in a more broad scenario not considering only the snap, but also other distros packages and security policies).

Even that, for any other reasons (let's say, just because they just don't like it), they decide to remove this function, we can easily patch putting it back again, since the function consists of two syscalls operations using the [sys/unix internal go library](https://pkg.go.dev/golang.org/x/sys/unix):

```go
func realChroot(path string) error {
	if err := unix.Chroot(path); err != nil {
		return fmt.Errorf("error after fallback to chroot: %w", err)
	}
	if err := unix.Chdir("/"); err != nil {
		return fmt.Errorf("error changing to new root after chroot: %w", err)
	}
	return nil
}
```

> Can we maybe add a test case to catch this issue if it becomes a problem in the future?

I've added the regression test here: 
- https://github.com/canonical/docker-snap/pull/285

locnnil:
The last run shows that the patch is working properly:

https://github.com/canonical/docker-snap/actions/runs/15610495632?pr=284

jpm-canonical (dismissed):

patches/engine/0006-use-realchroot-fallback-func.patch
```diff
@@ -0,0 +1,33 @@
...
++++ b/vendor/github.com/moby/go-archive/internal/mounttree/switchroot_linux.go
+@@ -23,6 +23,15 @@ func SwitchRoot(path string) error {
+ 		}
+ 	}
+ 
++	// See: https://github.com/canonical/docker-snap/issues/281
++	// During the container extraction layers, if any image has
++	// or creates the `/snap` directory, due to the pivot_root(2)
++	// syscall, the aa profile `snap.docker.dockerd` end up being
++	// activated on the 'new' root environment.
```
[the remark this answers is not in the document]
locnnil: AppArmor doesn't 'deny syscalls', this is an operation done by `seccomp(2)`, it actually checks (against rules define in a profile) controlling access to system resources. The denial of the syscalls related to the creation of `/snap` is a side effect of how it operates.

But I've added your suggestions here: https://github.com/canonical/docker-snap/pull/284/commits/d11d2dd0bd13e9c26880b49aef3188cc7d03dfae

.github/workflows/smoke-test.yml
```diff
@@ -200,6 +200,13 @@ jobs:
           sudo docker run --rm debian-hello
           sudo docker image rm debian-hello
 
+      # NOTE: regression test for check patch that forces
```
jpm-canonical, line 203: Indentation looks weird
```suggestion
       # NOTE: regression test for check patch that forces
```

patches/engine/0006-use-realchroot-fallback-func.patch
```diff
@@ -0,0 +1,36 @@
...
+ 	}
+
++	// This patch forces the use of the chroot(2) instead of pivot_root(2).
++	//
++	// During a docker pull, AppArmor denies operations related to the creation
++	// of /snap directory in an image layer.
++	// This is resulted by the pivot_root(2) syscall perfomed durnig
++	// the extraction of the image layer, confusing AppArmor to enforce
++	// the rules of snap.docker.dockerd profile on the wrong root environment.
++	// See: https://github.com/canonical/docker-snap/issues/281
```
jpm-canonical, line 26:
```suggestion
+	// This patch forces the use of the chroot(2) instead of pivot_root(2).
+	//
+	// During a docker pull, AppArmor denies operations related to the creation
+	// of a /snap directory in an image layer.
+	// This is caused by the pivot_root(2) syscall performed during
+	// the extraction of the image layer, confusing AppArmor to enforce
+	// the rules of snap.docker.dockerd profile on the wrong root environment.
+	// See: https://github.com/canonical/docker-snap/issues/281
```

jpm-canonical (changes requested):
Any idea why the tests are failing?

farshidtz:
Is the other existing patch making a similar change still required: https://github.com/canonical/docker-snap/blob/ac53baeb525611ab9d70178ed1b3e7a5229e8d1f/patches/engine/0005-snappy-real-chroot.patch

locnnil:
@jpm-canonical The patch was failing because I was naively trying to modify it. You cannot just edit the patch (at least not if you're adding/deleting lines), because the number of lines modified might change, and then the patch will be broken.

PR #300, opened 2025-07-07
ci: Add snap refresh tests for smoke-tests and Nvidia

locnnil:


.github/workflows/smoke-test.yml
```diff
@@ -104,6 +104,43 @@ jobs:
 
           set -e
 
+      - name: Snap refresh test
+        run: |
+
+          trap 'echo "error, sad day ($?)"; sleep 1; sudo snap logs -n=40 docker.dockerd; sleep 1; sudo tail -n20 /var/log/*.log; sudo dmesg | tail -n20; sudo journalctl --no-pager | grep docker' ERR
+          sudo snap install docker
+
+          sleep 5 # Give time for the daemon to startup
```
farshidtz, line 113:
```suggestion
          # Give time for the daemon to startup
          sleep 5
```

farshidtz (commented):
Looks good. I've posted some early feedback.

.github/workflows/smoke-test.yml
```diff
@@ -104,6 +104,43 @@ jobs:
...
+
+          # Run well known application with tagged versions
+          sudo docker run --restart=always --detach -p 6379:6379 redis:8.0.2
+          sudo docker run --restart=always --detach -e RABBITMQ_DEFAULT_USER=user -e RABBITMQ_DEFAULT_PASS=pass rabbitmq:3-management
+          sudo docker run --restart=always --detach -p 3000:3000 grafana/grafana:12.0.2-ubuntu
+          sudo docker run --restart=always --detach -p 8080:8080 -p 50000:50000  jenkins/jenkins:jdk21
+          sudo docker run --restart=always --detach -p 7070:80 nginx:1.29.0
+          sudo docker run --restart=always --detach -e POSTGRES_USER=user -e POSTGRES_PASSWORD=pass -e POSTGRES_DB=db postgres:16-alpine
```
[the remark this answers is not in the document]
locnnil: Added 28e56c5

.github/workflows/smoke-test.yml
locnnil: Fixed here: bfb5c6a

.github/workflows/smoke-test.yml
```diff
@@ -104,6 +104,49 @@ jobs:
...
+          sudo snap install --dangerous ./docker_${{ github.run_id}}_${{ matrix.runner.arch }}.snap
+
+          sleep 10
+
+          export UP_CONTAINERS_NEW=$(sudo docker ps -q | wc -l)
+
+          if [ "$UP_CONTAINERS_NEW" -lt "$UP_CONTAINERS_OLD" ]; then
+            echo "Error: not all containers started after snap refresh"
+            sudo docker ps -a
+            for i in $(sudo docker ps -q --filter "status=exited"); do
```
locnnil, line 140: Looking at the sources, these are the possible container status:

https://github.com/moby/moby/blob/1de9f39f2c193cb6738d4/api/types/container/state.go#L14-L20

So, should we also seek for `dead`? My opinion is that it wouldn't be necessary (since those are the ones which failed to be deleted).

PS: There is also the mention on [other places](https://github.com/search?q=repo%3Amoby%2Fmoby+OOMKilled+language%3AGo+&type=code) of a state called `OOMKilled` which means that it got killed by the [out-of-memory killer of Linux](https://docs.kernel.org/admin-guide/mm/concepts.html#oom-killer).

locnnil: So, should I just drop the acquisition of the logs of the containers who exited? Because I think it's pretty handy to have this upfront.

PR #305, opened 2025-07-15
feat: extend refresh test with snap revert

locnnil:
- Add revert test to complement the refresh test: Revert back to the original version after the refresh is performed.


.github/workflows/smoke-test.yml
```diff
@@ -120,26 +120,39 @@ jobs:
...
-
           # Refresh Docker snap to the current build
           sudo snap install --dangerous ./docker_${{ github.run_id}}_${{ matrix.runner.arch }}.snap
 
           sleep 10
 
           export UP_CONTAINERS_NEW=$(sudo docker ps -q | wc -l)
 
-          if [ "$UP_CONTAINERS_NEW" -lt "$UP_CONTAINERS_OLD" ]; then
+          if [ "$UP_CONTAINERS_NEW" -ne "$UP_CONTAINERS_OLD" ]; then
```
farshidtz, line 137:
```suggestion
          if [ "$RUNNING_CONTAINERS" -ne "$STARTED_CONTAINERS" ]; then
```

.github/workflows/smoke-test.yml
```diff
@@ -120,26 +120,39 @@ jobs:
           sudo docker run --restart=always -d jenkins/jenkins
           sudo docker run --restart=always -d nginx
           sudo docker run --restart=always -d -e POSTGRES_HOST_AUTH_METHOD=trust postgres
+          # NOTE: adjust this accordingly with the commands above if needed
+          UP_CONTAINERS_OLD=6
 
           # Wait for the restart policy to take effect
           # See: https://docs.docker.com/engine/containers/start-containers-automatically/#restart-policy-details
           sleep 10
```
farshidtz, line 128: Add
```
echo "List of running containers"
docker ps -a
```
and possibly a check to know if a failure was caused by normal container startup error or the subsequent refresh.

farshidtz: The trap just prints logs in case of a failure. Nothing is checked there.

.github/workflows/smoke-test.yml
```diff
@@ -120,26 +120,47 @@ jobs:
...
+          echo "List of running containers"
+          docker ps -a
+
+          if [[ -n $(sudo docker ps -q --filter "status=exited") ]]; then
+            echo "Error: docker snap did not start all containers"
+            exit 1
+          fi
+
+          # NOTE: adjust this accordingly with the commands above if needed
+          STARTED_CONTAINERS=6
```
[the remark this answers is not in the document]
locnnil: True, it should be before the echo listing the running containers.

farshidtz (approved):

PR #317, opened 2025-09-11
Upgrade v28.4.0 (core22)

locnnil:
- Cherry picking `v28.4.0` upgrade from main **core24** based snap to **core22** based snap.


locnnil:
This cherry-pick end up being very messy. There is a lot of core24 related diff around.

Issue #138, opened 2023-05-23
Somehow detect and use xtables on focal


xnox:
This is inverse of https://github.com/docker-snap/docker-snap/issues/68

I don't know if for example we have access to hostfs and if we can sniff iptables command from there somehow.

locnnil:
Issue motivation on this comment: https://github.com/canonical/docker-snap/issues/68#issuecomment-1558842511

fredldotme:
I've implemented a way to have both nftables and xtables in the snap, and let the user select the appropriate one: https://github.com/canonical/docker-snap/pull/327

Issue #68, opened 2022-04-11
docker snap unconditionally uses xtables


jdstrand:
https://bugs.launchpad.net/ufw/+bug/1968608 has a great deal of information on the background of the problem, but in essence, the kernel supports two different firewall interfaces: xtables (historic) and netfilter (modern). Per kernel upstream, all software on the system must agree to all use xtables or all use netfilter, otherwise undefined behavior can occur. For docker, this might be that containers are unable to obtain an ip address, etc.

Distributions like Ubuntu 22.04 are migrating to using netfilter as the default kernel interface, but the docker snap unconditionally uses xtables. This can be seen on an up to date Ubuntu 22.04 system. Eg:

```
$ sudo ufw allow 22 && sudo ufw enable
$ sudo snap install lxd
$ sudo snap install multipass
$ sudo iptables-nft -S
... lots of rules ...
$ sudo iptables-legacy -S
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
```

In the above, the `iptables-legacy -S` output shows that nothing is using the xtables backend. Install the docker snap, and it does:
```
$ sudo snap install docker
$ sudo iptables-legacy -S
-P INPUT ACCEPT
-P FORWARD DROP
-P OUTPUT ACCEPT
-N DOCKER
-N DOCKER-ISOLATION-STAGE-1
-N DOCKER-ISOLATION-STAGE-2
-N DOCKER-USER
-A FORWARD -j DOCKER-USER
-A FORWARD -j DOCKER-ISOLATION-STAGE-1
-A FORWARD -o docker0 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
-A FORWARD -o docker0 -j DOCKER
-A FORWARD -i docker0 ! -o docker0 -j ACCEPT
-A FORWARD -i docker0 -o docker0 -j ACCEPT
-A DOCKER-ISOLATION-STAGE-1 -i docker0 ! -o docker0 -j DOCKER-ISOLATION-STAGE-2
-A DOCKER-ISOLATION-STAGE-1 -j RETURN
-A DOCKER-ISOLATION-STAGE-2 -o docker0 -j DROP
-A DOCKER-ISOLATION-STAGE-2 -j RETURN
-A DOCKER-USER -j RETURN
```

The docker snap should've instead added rules to netfilter. The ufw, lxd and multipass snaps all detect when to use netfilter vs xtables. The docker snap should do the same.

[...]

PR #327, opened 2025-11-06
bin & snap: Automatic detection between nftables and legacy xtables

fredldotme:
Introduces iptables wrappers to detect and select whether to use nftables
or legacy xtables for configuring the firewall. This is done by running
a simple 'nft list ruleset' command to see whether it succeeds or not.
If it does not, fall back to using iptables-legacy as symlinked in the
snap, instead of using the wrong symlink from core24.

Addresses: https://github.com/canonical/docker-snap/issues/138

### DIFF
--- a/bin/ip6tables
+++ b/bin/ip6tables
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+# If kernel support is missing, automatically assume xtables
+USE_NFTABLES="$(nft list ruleset > /dev/null 2> /dev/null && echo true || echo false)"
+
+# Otherwise resolve the iptables symlink from the host and use
+# xtables if it is pointing to xtables-legacy-multi
+if [ "$USE_NFTABLES" == "true" ]; then
+    if readlink -f /var/lib/snapd/hostfs/sbin/ip6tables 2>/dev/null | grep -q xtables-legacy-multi; then
+        USE_NFTABLES=false
+    fi
+
+    # If it's not a symlink then it is pure a pure xtables implementation
+    if readlink -f /var/lib/snapd/hostfs/sbin/ip6tables 2>/dev/null | grep -q "^/var/lib/snapd/hostfs/sbin/ip6tables"; then
+        USE_NFTABLES=false
+    fi
+fi
+
+if [ "$USE_NFTABLES" == "true" ]; then
+    IMPL=ip6tables-nft
+else
+    IMPL=ip6tables-legacy
+fi
+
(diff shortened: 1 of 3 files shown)


farshidtz (commented):
Thanks a lot for your contribution. 

The Docker Engine is adding a daemon-level configuration for the firewall backend configuration in v29.0.0, with a new `firewall-back` configuration field supporting [`nftables` and `iptables`](https://github.com/moby/moby/blob/d166f42ef29ea87fb946772619b4c4a4968f9f56/daemon/config/config_linux.go#L303) as input. See:  https://docs.docker.com/engine/network/firewall-nftables/

Because of the above, I'm not keen on adding a similar configuration to the snap as it could conflict with what is set as daemon configuration. 

As explained in https://github.com/canonical/docker-snap/issues/68#issuecomment-1487348861, the root cause of this issue is that the snap base ships the unwanted symlinks. A perhaps simpler workaround could be to detect the firewall backend (using `iptables --version`) and automatically execute the right command. This will alleviate the need for a new snap configuration. 

On Ubuntu 20.04:
```console
$ iptables --version
iptables v1.8.4 (legacy)

$ sudo snap run --shell docker.dockerd
# realpath /usr/sbin/iptables
/usr/sbin/xtables-nft-multi
# iptables --version
iptables v1.8.10 (legacy)
```

On Ubuntu 24.04:
```console
$ iptables --version
iptables v1.8.10 (nf_tables)

$ sudo snap run --shell docker.dockerd
# realpath /usr/sbin/iptables
/usr/sbin/xtables-nft-multi
# iptables --version
iptables v1.8.10 (nf_tables)
```

I've released a snap with your additions to `latest/edge/pr-327` channel for experimental use, while waiting for a snap release of v29.0.0. The channel should stay available for the next 30 days, and we can extend it if v29 doesn't become available in time.

fredldotme:
Update: I tried to reproduce your automatic detection sample, and it fails on Ubuntu Touch:

```
phablet@ubuntu-phablet:~$ iptables --version
iptables v1.8.10 (legacy)
phablet@ubuntu-phablet:~$ sudo snap run --shell docker.dockerd
root@ubuntu-phablet:/home/phablet# iptables --version
iptables v1.8.10 (nf_tables)
root@ubuntu-phablet:/home/phablet#
```

I instead opted for checking for a successful `nft list ruleset` execution, and fall back to iptables-legacy in case of an error.

fredldotme:
I suppose `xtables-multi` is pure xtables without nft support, in which case I would have to add that to the `iptables` wrappers in the other PR.

PR #328, opened 2025-11-11
feat: add content interface to write configurations from another snap

### LINKED ISSUE #299: content slot for rw access to docker daemon config
### What needs to get done

We have a brandstore customer request for rw to daemon.json from their configuration snap. They already have snapd-control and shutdown access so the limitations discussed in #158 shouldn't be an issue. Would it be possible to provide so? I expect other customers to come up with same request. 

### Why it needs to get done

We discussed about this internally with snapd team and came to conclusion that this would be recommended solution because:
- Using system-interface isn't ideal solution as you need to give full path for it and thus it would be bound to a specific revision only. 
- Wildcards aren't applicable because system-files won't allow * and apparmor won't understand .+ 
- Giving access to /var/snap/docker would be too greedy.

farshidtz:
Adding the write-slot that was excluded from https://github.com/canonical/docker-snap/pull/158

Closes https://github.com/canonical/docker-snap/issues/299

The snap that writes the config needs to have a corresponding plug, e.g.:
```yaml
plugs:
  docker:
    interface: content
    content: docker-config
    target: $SNAP_DATA/docker-config
```
And a logic that modifies the config, e.g. `snap/hooks/connect-plug-docker`:
```bash
# Set log-level to debug
CONFIG_FILE="$SNAP_DATA/docker-config/daemon.json"  
jq '.["log-level"] = "debug"' "$CONFIG_FILE" \
    > "$CONFIG_FILE.tmp" && \
    mv "$CONFIG_FILE.tmp" "$CONFIG_FILE"
``` 

Full example: https://github.com/farshidtz/docker-config-provider-snap

> [!CAUTION] 
> Changes made to daemon.json may be overwritten. The Docker snap with nvidia support writes to the config directory and modifies the daemon.json file during the setup.


lczyk:
blocked on:

- https://github.com/canonical/docker-snap/pull/384

lczyk (changes requested):
hmm... i dont really like the current solution as is. snapd executes plugs concurrently (as best as i can tell from the snapd source anyway: plug tasks run [here](https://github.com/canonical/snapd/blob/1a2152a27d0c9d09927d68a881c6f1ba8b61e688/overlord/state/taskrunner.go#L383) and [this](https://github.com/canonical/snapd/blob/1a2152a27d0c9d09927d68a881c6f1ba8b61e688/overlord/state/taskrunner.go#L477) launches each as a concurrent goroutine) and in undetermined order so there is a real and stochastic race between all the plugs *and* `nvidia-container-toolkit`.

i'd at the very least propose an `flock` on daemon.json but that relies on the plugs also cooperating and using the flock, but at least there is a way for the plugs not to get clobbered by other writers if all are cooperating. atm there literally isnt a way.

---

a more complicated alternative could be something like this:

```
$SNAP_DATA/config/
  daemon.json.d/
    10-nvidia.json
    20-foobar.json
```

and then merging them into daemon.json in `bin/dockerd-wrapper` with something like `jq -s 'reduce .[] as $x ({}; . * $x)' config/daemon.json.d/*.json > daemon.json`. i like this solution much more personally, but it would be w bigger change because it would also need to touch what `nvidia-container-toolkit` is doing, although not too much.

farshidtz:
@lczyk Thanks for your review.

As the PR description suggests, this method of providing configurations is not compatible with when nvidia-container-toolkit is used. I do like the idea of making it compatible and suggest creating a separate PR with your proposal, in favor of this PR. 

Regardless, I believe snap auto-connections do happen deterministically compared to the startup of services. When the Docker snap is manually installed after the config provider, the connection happens before services start. On Ubuntu Core (where this is expected to be used), the connections and service startups happen in separate stages.

farshidtz:
Superseded by https://github.com/canonical/docker-snap/pull/391

PR #329, opened 2025-11-13
xtables fallback

fredldotme:
Support both nftables as well as xtables in various configurations.

Support three use-cases:
- Use iptables-legacy should the kernel not support nftables
- Use iptables-legacy if it supports nftables but the user's host
  has a symlink to xtables-legacy-multi
- User configuration via $SNAP_DATA/config/daemon.json

This is needed as otherwise systems with only xtables in use still won't work with v29.0.0.

Fixes: https://github.com/canonical/docker-snap/issues/138

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -297,27 +297,46 @@ jobs:
       - name: Docker using xtables-nft
         run: |
           sudo snap stop docker
+
           sudo nft flush ruleset
+          set +e
+          COUNT="$(sudo nft list ruleset 2>/dev/null | grep -i docker | wc -l)"
+          set -e
+          [ "$COUNT" -eq 0 ]
 
           sudo update-alternatives --set iptables /usr/sbin/iptables-nft
           sudo update-alternatives --set ip6tables /usr/sbin/ip6tables-nft
 
           sudo snap start docker
-          sleep 5
+          sleep 10
 
+          set +e
           COUNT="$(sudo nft list ruleset 2>/dev/null | grep -i docker | wc -l)"
+          set -e
           [ "$COUNT" -gt 0 ]
 
       - name: Docker using xtables-legacy
         run: |
           sudo snap stop docker
+
           sudo nft flush ruleset
+          set +e
+          COUNT="$(sudo nft list ruleset 2>/dev/null | grep -i docker | wc -l)"
+          set -e
+          [ "$COUNT" -eq 0 ]
 
           sudo update-alternatives --set iptables /usr/sbin/iptables-legacy
           sudo update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 
-          sudo snap start docker
-          sleep 5
-
+          set +e
           COUNT="$(sudo iptables -L | grep -i docker | wc -l)"
+          set -e
+          [ "$COUNT" -eq 0 ]
+
+          sudo snap start docker
+          sleep 10
+
+          set +e
+          COUNT="$(sudo iptables -L | grep -i docker | wc -l)"
+          set -e
           [ "$COUNT" -gt 0 ]


bin/ip6tables
```diff
@@ -0,0 +1,25 @@
...
+    fi
+fi
+
+if [ "$USE_NFTABLES" == "true" ]; then
+    IMPL=ip6tables-nft
+else
+    IMPL=ip6tables-legacy
+fi
+
+exec "$SNAP/usr/sbin/$IMPL" "$@"
```
farshidtz: Why don't we call the `xtables-nft-multi`/`xtables-legacy-multi` commands from the host or staged debs directly?

bin/iptables
```diff
@@ -0,0 +1,25 @@
...
+    fi
+fi
+
+if [ "$USE_NFTABLES" == "true" ]; then
+    IMPL=iptables-nft
+else
+    IMPL=iptables-legacy
+fi
+
+exec "$SNAP/usr/sbin/$IMPL" "$@"
```
farshidtz: Can we avoid the extra symlinks altogether?
```suggestion
if [ "$USE_NFTABLES" == "true" ]; then
    IMPL=xtables-nft-multi
else
    IMPL=xtables-legacy-multi
fi

exec "$SNAP/usr/sbin/$IMPL" iptables "$@"
```

fredldotme:
> > What system was that on? I would like to reproduce this.
> 
> That was on Ubuntu Core 18, in a multipass VM. Find all available images with `multipass find`.
> 
> ```
> multipass launch --name uc18 core18
> multipass shell uc18
> ```
> 
> ```
> $ uname -r
> 4.15.0-243-generic
> ```
> 
> I don't get the firewall rules with Docker 29.0.0 from the edge channel, either. But the Docker engine works in all cases (this PR, edge, stable).

I suppose it was the missing check for `xtables-multi` on 18. I've test-installed the latest branch state and got this:

```
ubuntu@uc18:~$ sudo snap run --shell docker.dockerd
root@uc18:/home/ubuntu# nft list ruleset
root@uc18:/home/ubuntu# bash -x /snap/docker/x1/bin/iptables -L
++ nft list ruleset
++ echo true
+ USE_NFTABLES=true
+ '[' true == true ']'
+ grep -q 'xtables-legacy-multi$'
+ readlink -f /var/lib/snapd/hostfs/sbin/iptables
+ grep -q 'xtables-multi$'
+ readlink -f /var/lib/snapd/hostfs/sbin/iptables
+ USE_NFTABLES=false
+ grep -q '^/var/lib/snapd/hostfs/sbin/iptables$'
+ readlink -f /var/lib/snapd/hostfs/sbin/iptables
+ '[' false == true ']'
+ BIN=xtables-legacy-multi
+ exec xtables-legacy-multi iptables -L
Chain INPUT (policy ACCEPT)
target     prot opt source               destination         

Chain FORWARD (policy DROP)
target     prot opt source               destination         
DOCKER-USER  all  --  anywhere             anywhere            
DOCKER-FORWARD  all  --  anywhere             anywhere            

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination         

Chain DOCKER (1 references)
target     prot opt source               destination         
DROP       all  --  anywhere             anywhere            

Chain DOCKER-BRIDGE (1 references)
target     prot opt source               destination         
DOCKER     all  --  anywhere             anywhere            

Chain DOCKER-CT (1 references)
target     prot opt source               destination         
ACCEPT     all  --  anywhere             anywhere             ctstate RELATED,ESTABLISHED

Chain DOCKER-FORWARD (1 references)
target     prot opt source               destination         
DOCKER-CT  all  --  anywhere             anywhere            
DOCKER-INTERNAL  all  --  anywhere             anywhere            
DOCKER-BRIDGE  all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            

Chain DOCKER-INTERNAL (1 references)
target     prot opt source               destination         

Chain DOCKER-USER (1 references)
target     prot opt source               destination
```

@locnnil will do, might take a few attempts until I get it right. Will ping you once it's done.

locnnil:
Changes looks good, thanks!

I've tested on:

- Fedora Cloud 43 (amd64)  (`uname -r`:  6.17.8-300.fc43.x86_64) - distro uses nftables
- ~~Fedora Cloud 31 (amd64) (the last version that used to have iptables by default)~~ (already on EOL, do doesn't matter)
- ArchLinux (amd64) (`uname -r`: 6.17.9-arch1-1) - distro uses iptables 
- CentOS Stream Generic Cloud 9 (amd64) (`uname -r`: 5.14.0-642.el9.x86_64) - distro uses nftables
- Raspberry Pi OS Lite (Debian GNU/Linux 13 (trixie)) (arm64) (`uname -r`: 6.12.47+rpt-rpi-v8) - distro uses nftables

farshidtz:
> Should I also add a native nft test through configuring it in the the `daemon.json` or is it too soon to "support" something like this?

Thanks, we can live with that, since this is still experimental:

> Support for nftables introduced in Docker 29.0.0 is experimental, configuration options, behavior and implementation may all change in future releases. The rules for overlay networks have not yet been migrated from iptables. Therefore, nftables cannot be enabled when the Docker daemon is running in Swarm mode. [[1](https://docs.docker.com/engine/network/firewall-nftables/)]

PR #330, opened 2025-11-13
Upgrade to v29.0.0

fredldotme:
Refreshes patches and gets rid of the GO111MODULE hack in the meantime.


farshidtz:
Triggered tests for the nvidia runtime: https://github.com/canonical/docker-snap/actions/runs/19331957284

PR #349, opened 2026-02-04
CI: Fix NVIDIA test on core24

jpm-canonical:
There is an issue with the kernel components on Ubuntu Core 24. The NVIDIA drivers v570 and v580 fail to install. v550 installs but does not provide any content. After a discussion with the Core and Snapd team, we were recommended to upgrade to Snapd v2.74, as that fixes a bug that prevents the installation of the NVIDIA components.

This PR improves the checking and retrying of snap refreshes at startup, and adds retries for snap and apt installs. This should improve reliability of the tests on lab hardware with intermittent network issues.

Some further changes are made to the testflinger scripts to make local testing easier.

Example run: https://github.com/canonical/docker-snap/actions/runs/21746070241

Also tested #344 in this run: https://github.com/canonical/docker-snap/actions/runs/21749855502


farshidtz (approved):
Thanks for the fixes and improvements 🧡

PR #369, opened 2026-04-09
feat(ci): retry Docker Hub pulls + pin tags in smoke-test refresh step

danielnovais-tech:


danielnovais-tech:
@copilot apply changes based on the comments in [this thread](https://github.com/canonical/docker-snap/pull/369#pullrequestreview-4086058728)

lczyk:
hi @danielnovais-tech,

sorry to get to this so late, we've had some reshuffle of maintainership (hi 👋)

what was the original purpose of this PR? was it some dockerhub transient 500's? i don't really see any failures in the last 4 months or so that this PR would help with.

also, if it's unauth'ed dockerhub pull limits that are the issue, i'd say a much better solution would be to auth. atm though there is nothing which makes me think we need either

PR #371, opened 2026-04-21
Always include architecture suffix in spread systems

imatrisciano:
This PR adds the architecture suffix to every test system in `spread.yaml`.

The main purpose of this change is to make it easier to target a single architecture by using a filter when running spread tests, for example, you can now run x86_64 tests only by executing `spread ubuntu...amd64`. 

Please note that _spread_ does not allow using the character `_` in system names, so the `x86_64` architecture suffix cannot be included in the system name. To target the `x86_64` architecture, the  `amd64` suffix is used instead. For consistency, the `arm64` suffix is used to target the `aarch64` architecture.

### DIFF
--- a/spread.md
+++ b/spread.md
@@ -52,10 +52,16 @@ The system will download the virtual machine files and place them in the `.image
 
 To save time you can select a subset of systems and tests to run.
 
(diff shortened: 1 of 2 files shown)


farshidtz (approved):
Thanks. I particularly like the use of Debian architecture strings as it adds consistency with architecture values used for snap artifacts passed as input.

PR #372, opened 2026-04-21
feat(ci): improve spread tests resilience

imatrisciano:
Spread tests may fail in slower systems due to time-related issues, especially if emulating a different architecture.

This PR tries to improve tests resilience by:

- increasing the task timeout
- increasing the provisioning wait time if KVM is not available
- converting tests to use a retry logic instead of blindly rely on sleeping
- retrying image-garden allocations on failure

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -21,6 +21,12 @@ on:
   # Manual trigger
   workflow_dispatch:
 
+# Cancel an in-progress run when a newer commit lands on the same ref, so a
+# burst of pushes to a branch or PR only tests the latest one.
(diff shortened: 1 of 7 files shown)


lczyk:
thanks @imatrisciano! this is really quite great! i've been trying to make the arm tests work better and this basically does half of it already! i've done a small tweak here (see below), and will do a followup PR with other things i found.

### tweak

the retry wrapped an inline command substitution, so it re-checked a frozen value on every attempt -- moved it into a function so `docker inspect` actually re-runs each retry

lczyk (dismissed):

lczyk (approved):

PR #388, opened 2026-07-21
fix(ci): reliable emulated spread runs

lczyk:
builds on #372.

emulated cross-arch spread runs were effectively unusable: a VM that never became
reachable hung the allocation until github's 6h ceiling, so a run would burn 6h and
execute zero tasks. this PR makes tests much more boring: 23 systems across 5 architectures, one job each, bounded, cached, and filterable.

### one system per job

`list-systems` job expands the job filter into concrete spread jobs and fans them
out as a dynamic matrix: one runner per system, one qemu guest each. guests no longer
compete for a single runner's cpu and memory, so the boot-contention failure modes
(udev starvation dropping guests into emergency mode, host oom kills) are gone, and systems run in parallel -- the full board finishes reliably in under an hour.

the filter is a chain of whitespace-separated expressions applied as successive greps over
the `spread -list` output: each term narrows, `!` inverts a term, `|` inside a term
gives alternatives, an empty filter selects everything, and task names work too
(`!26.04 cloud amd64 build`). an ubuntu core release counts as part of its paired
LTS series: `26.04` also selects core-26, `!26.04` also drops it, `26.10` leaves
core alone. the expansion lives in `.github/scripts/spread-matrix` (bats-tested).

### the platter

- core-22/24 on amd64 and arm64
- core-26 on amd64 and arm64 **(new)**
- cloud 22.04 on amd64 and arm64
- cloud 24.04/26.04 on amd64, arm64
- cloud 24.04/26.04 on riscv64 **(not new but was also v broken)**
- cloud 24.04/26.04 on s390x and ppc64el **(new)**
- cloud 26.10 aka devel, on all five arches **(new)**. expected to fail with the current stable image-garden but works with `latest/candidate` (v0.6.4)
- no core for riscv64/s390x/ppc64el: ubuntu core is not published for these arches

### bounded allocation, fail fast

`image-garden allocate` blocks until the guest is ssh-reachable and nothing bounds it when a guest never gets there. `timeout` isn't shipped in the `image-garden` snap and the host binary is blocked by confinement, so the allocate script carries its own small watchdog (30m), then fails the allocation with `FATAL` instead of hanging. allocation also:

- polls for the actual ssh banner on emulated systems -- image-garden hands the address out once the port-forward exists. this is in a race with sshd actually picking up the phone. normally not an issue but under emulation everything works in slo-motion and the wall-time gap between port-forward and sshd gets exaggerated.
- deletes the half-built qcow2 overlay on failure, so the cache can never capture an unprovisioned image
- runs under an explicit backend-level `kill-timeout` -- spread's project-level setting does not cascade to adhoc allocate scripts, which otherwise get the 15m default; too short for a cold emulated ubuntu-core provision boot

### firmware and machine quirks

image-garden snap packaging issues worked around in `spread/scripts/prepare-firmware.sh` and `allocate.sh` (upstream reports to follow):

- **aarch64**: the emitted EFI vars image is ~2.8 MB but qemu's virt machine requires exactly 64 MiB (`cfi.pflash01 device '/machine/virt.flash1' requires 67108864 bytes`); repaired up front with `qemu-img resize`, strictly scoped to aarch64 (x86_64's q35 caps firmware at 8 MiB and must not be grown); if the image still has the wrong size after the repair, allocation fails immediately with a clear error instead of an opaque boot hang
- **riscv64**: image-garden pads its flash images with `truncate` (which the snap neither ships nor may execute from the host btw.. 😑 ) correctly sized images are pre-built with `qemu-img` instead
- **ppc64el**: the qemu-ppc64 component doesn't ship `vgabios-stdvga.bin`, which the pseries machine wants for its default VGA; video is disabled outright (`-nographic -vga none`)

### ubuntu core first boot

core images seed from an old recovery system, so the first boot auto-refreshes the boot snaps and reboots -- mid-prepare, aborting every task (`snapd is about to reboot the system`). `prepare.sh` holds refreshes for the lifetime of the ephemeral guest.

### caching provisioned images

explicit `actions/cache` restore/save split so provisioned images persist after failed runs too (but not cancellations, which can leave a half-written overlay). one cache entry per system, keyed by the `image-garden` snap revision. a prefix restore from an older revision keeps the downloads warm but drops the provisioned overlays, so a revision bump re-provisions from scratch instead of trusting stale images.

### dispatch inputs

- `snap_channel` -- `docker` channel to test
- `spread_test_filter` -- see above
- `force_rebuild` -- ignore cached vm images
- `publish_boot_logs` -- always upload guest boot logs (uploaded on failure
  regardless)
- `debug` -- spread `-vv`
- `image_garden_channel` -- `image-garden` snap channel

### tests

the fiddly shell bits are extracted into sourceable/executable scripts with bats coverage: the matrix builder (`.github/scripts/spread-matrix`), the allocation watchdog and the ssh readiness poll (`spread/scripts/`). the new `scripts-tests` workflow runs bats plus shellcheck on PRs touching workflows or scripts.

validated with manual dispatches:

- full board on stable: https://github.com/canonical/docker-snap/actions/runs/29908829222
  -- 18/18 supported lanes green (26.10 red as expected on stable); amd64 3-5m,
  arm64 13-26m, ppc64el 15-20m, riscv64 28-33m, s390x 31-47m
- 26.10 on candidate: https://github.com/canonical/docker-snap/actions/runs/29909392492
  -- 5/5 green, cold, 5-42m


lczyk:
~~dropping this to draft for a sec. arm was being very slow on cold cache and i think i have a better + simpler idea too. wip 🚧~~ done

spread.yaml
```diff
@@ -1,5 +1,8 @@
 project: docker
-kill-timeout: 3h
+# Headroom for slow, emulated cross-architecture boots. Failures are bounded
+# well below this by the per-attempt timeout and FATAL in the allocate script,
+# so this only ever caps a genuinely slow (but progressing) boot.
+kill-timeout: 45m
```
dariuszd21, line 5: This is the same value as in `backends:`.
I don't remember by heart, but can it race in the spread world?

spread.yaml
```diff
@@ -12,63 +15,15 @@ backends:
...
-      # original path before such modifications were applied. Snap
-      # applications cannot normally run /snap/bin/* entry-points
-      # successfully so re-set PATH to the original value, as provided by
-      # snapcraft.
-      if [ -n "${SPREAD_HOST_PATH-}" ]; then
-        PATH="${SPREAD_HOST_PATH}"
-      fi
-
-      image-garden discard "$SPREAD_SYSTEM_ADDRESS"
+    allocate: source spread/scripts/allocate.sh
```
dariuszd21, line 25: nit: as I've mentioned during our sync, I've seen calling the script directly in such code rather than sourcing it (and that is also what `craft test` [does](https://github.com/canonical/craft-application/blob/b67f6ca6a57e6ca15b16c8a629622741b1df5066/craft_application/services/testing.py#L264-L274) ).

spread.yaml
```diff
@@ -119,7 +74,29 @@ backends:
           password: ubuntu
           environment:
             SNAP_FILE: $(HOST:echo $SNAP_FILE_RISCV64)
+      # No ubuntu-core systems for s390x/ppc64el: Ubuntu Core is not published
+      # for those architectures.
```
dariuszd21, line 78: I don't think this is the best place for this comment (above the `-cloud` system of 24? only).
It feels to me to be in a middle of nowhere.
Maybe before core/cloud we could just note which architectures are expected to be tested

spread.md
```diff
@@ -35,11 +39,11 @@ Before running any test, you have to choose which docker snap to test
   SNAP_CHANNEL=latest/edge image-garden.spread
   ```
 
-  This runs the tests for both amd64 and arm64 architectures.
+  This runs the tests on every system, each guest installing the snap for its own architecture from the store.
   For testing a single architecture, download the snap and use the local snap file method.
   To download the snap on a different architecture, e.g. arm64 on amd64, run: `UBUNTU_STORE_ARCH=arm64 snap download docker`.
 
-- To test a local snap file, specify the `SNAP_FILE_AMD64`, `SNAP_FILE_ARM64` and/or `SNAP_FILE_RISCV64`:
+- To test a local snap file, specify the `SNAP_FILE_<ARCH>` variable for each architecture you run (`AMD64`, `ARM64`, `RISCV64`, `S390X`, `PPC64EL`):
```
dariuszd21, line 46: You've added it here but the example below still doesn't mention S390X and PPC64EL

.github/workflows/spread-tests.yml
```diff
@@ -5,67 +5,194 @@ on:
...
+        description: "`docker` snap channel to test"
         required: true
-        type: string
+        type: choice
+        options:
+          - latest/stable
+          - latest/candidate
+          - latest/beta
+          - latest/edge
+        default: latest/edge
```
dariuszd21: This change makes it impossible to test any revision outside of those 4 channels.
I'm not sure if this is intended

spread/scripts/wait-for-ssh.bats
```diff
@@ -0,0 +1,75 @@
...
+}
+
+@test "a non-ssh banner is not ready" {
+  start_listener $'HTTP/1.0 200 OK\r\n'
+  run wait_for_ssh 127.0.0.1 "$PORT" 3 0.1
+  [ "$status" -eq 1 ]
+}
+
+@test "an unreachable port times out" {
+  run wait_for_ssh 127.0.0.1 "$(free_port)" 3 0.1
```
dariuszd21, line 66: This relies on a race condition, as `free_port()` doesn't keep the port open and closes the socket immediately on exit.
Theoretically, between free_port  finishes and ssh connection something could re-use this connection

spread/scripts/allocate.sh
```diff
@@ -0,0 +1,72 @@
...
+# One emulated guest per runner (the CI matrix fans systems out into separate
+# jobs), so there is no contention to budget memory against.
+export QEMU_SMP_OPTION="-smp 4"
+export QEMU_MEM_OPTION="-m 3072"
+
+# Map spread system architecture
+SPREAD_SYSTEM=${SPREAD_SYSTEM/%amd64/x86_64}  # maps "amd64" to "x86_64"
+SPREAD_SYSTEM=${SPREAD_SYSTEM/%arm64/aarch64} # maps "arm64" to "aarch64"
+
+GARDEN_SYSTEM="${SPREAD_SYSTEM/-plus-/+}"
```
dariuszd21, line 23: Do we have any `system` with `-plus-` ?

spread/scripts/allocate.sh
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+# One emulated guest per runner (the CI matrix fans systems out into separate
+# jobs), so there is no contention to budget memory against.
+export QEMU_SMP_OPTION="-smp 4"
+export QEMU_MEM_OPTION="-m 3072"
+
+# Map spread system architecture
+SPREAD_SYSTEM=${SPREAD_SYSTEM/%amd64/x86_64}  # maps "amd64" to "x86_64"
+SPREAD_SYSTEM=${SPREAD_SYSTEM/%arm64/aarch64} # maps "arm64" to "aarch64"
```
dariuszd21, line 21: Are you sure you want to shadow/mutate spread environment variable like that?

spread/scripts/allocate.sh
```diff
@@ -0,0 +1,72 @@
...
+# successfully so re-set PATH to the original value, as provided by
+# snapcraft.
+if [ -n "${SPREAD_HOST_PATH-}" ]; then
+  PATH="${SPREAD_HOST_PATH}"
+fi
+
+# One emulated guest per runner (the CI matrix fans systems out into separate
+# jobs), so there is no contention to budget memory against.
+export QEMU_SMP_OPTION="-smp 4"
+export QEMU_MEM_OPTION="-m 3072"
```
dariuszd21, line 17: Should this be configurable somehow, or our runners are expected to be able to accommodate for this?

.github/scripts/spread-matrix/spread-matrix.bats
```diff
@@ -0,0 +1,115 @@
...
+  "$SCRIPT" "$1" <<<"$JOBS" 2>/dev/null
+}
+
+# The selected system names for a filter, one per line, matrix order.
+systems() {
+  matrix "$1" | jq -r '.include[].system'
+}
+
+@test "empty filter selects every system with every task" {
+  [ "$(matrix '' | jq '.include | length')" -eq 23 ]
```
dariuszd21, line 52: Does it accept bash syntax to avoid magic numbers ?
```suggestion
  [ "$(matrix '' | jq '.include | length')" -eq ${#systems[@]} ]
```

.github/scripts/spread-matrix/spread-matrix.bats
```diff
@@ -0,0 +1,115 @@
...
+  done
+}
+
+# The include-matrix for a filter, stderr diagnostics dropped.
+matrix() {
+  "$SCRIPT" "$1" <<<"$JOBS" 2>/dev/null
+}
+
+# The selected system names for a filter, one per line, matrix order.
+systems() {
```
dariuszd21, line 47: This collides with local variable in `setup`, which is a bit confusing for a reader

.github/scripts/spread-matrix/spread-matrix.bats
```diff
@@ -0,0 +1,115 @@
...
+
+@test "negated LTS term drops its paired core release" {
+  [ "$(systems '!26.04' | grep -c 'core-26\.')" -eq 0 ]
+  [ "$(systems '!26.04' | wc -l)" -eq 16 ]
+}
+
+@test "development series term leaves core alone" {
+  [ "$(systems '26.10' | grep -vc 'cloud-26\.10')" -eq 0 ]
+  [ "$(systems '26.10' | wc -l)" -eq 5 ]
+}
```
dariuszd21, line 90: In other tests positive pattern matching is used for grep.
maybe the first one should be just `grep -c 'core'`

.github/scripts/spread-matrix/spread-matrix
```diff
@@ -0,0 +1,53 @@
...
+# treat them as one series: '26.04' selects core-26 too, '!26.04' drops
+# it, and '26.10' leaves core alone. Greps therefore run over two
+# tab-separated copies of each line -- the real job name plus a copy with
+# the core version expanded to LTS form -- and the real name comes back
+# out of column 1.
+set -eu
+
+FILTER="${1-}"
+
+jobs="$(cat)"
```
dariuszd21, line 24: This is input of the script, which mutates to output in line 34, maybe those two should be distinguished

spread/scripts/wait-for-ssh.sh
```diff
@@ -0,0 +1,33 @@
...
+# proof of anything: qemu's slirp port-forward accepts connections as soon
+# as it exists, well before the guest's sshd is up. Returns 0 once the
+# banner is seen, 1 when the attempts run out. Checks first, sleeps after
+# a failed check only, so a ready daemon is picked up immediately.
+
+_ssh_banner_ready() {
+  (
+    exec 3<>"/dev/tcp/$1/$2" || exit 1
+    banner=""
+    IFS= read -r -t 10 -n 4 banner <&3 || true
```
dariuszd21, line 15: This has a hard-coded 10 seconds timeout.
Maybe it should be documented/specified somewere

spread/scripts/allocate.sh
```diff
@@ -0,0 +1,72 @@
...
+  # completed and the overlay is worth caching.
+  rm -f ".image-garden/${GARDEN_SYSTEM}.qcow2"
+  FATAL "image-garden could not allocate $SPREAD_SYSTEM"
+fi
+
+if [ "$HOST_ARCH" != "$SYSTEM_ARCH" ] || [ ! -e /dev/kvm ]; then
+  # image-garden hands out the address once the port-forward exists, which under
+  # emulation can be well before sshd actually answers. Poll for the banner and
+  # hand over as soon as the daemon responds, with a hard 10-minute upper bound
+  # (120 attempts, 5s apart).
```
dariuszd21, line 62: This calculation is wrong, as read has `10s` timeout so it's up to `15s` * 120 (30 minutes)

dariuszd21 (changes requested):
Did a first round, happy to reiterate once those things will be addressed

spread/scripts/allocate.sh
lczyk: fixed in c8f09b1

spread/scripts/wait-for-ssh.sh
lczyk: yep. good point. fixed in c8f09b1

.github/scripts/spread-matrix/spread-matrix
lczyk: 58e8047

.github/scripts/spread-matrix/spread-matrix.bats
lczyk: 👍 1ede8d5

.github/scripts/spread-matrix/spread-matrix.bats
lczyk: ab19fbf

spread/scripts/allocate.sh
lczyk: it shouldn't really be an issue since its one guest per runner now, but no reason to force them, so 874d7fe turns them into defaults.

spread/scripts/allocate.sh
lczyk: it's harmless in practice, but you're right, shadowing a spread variable feels meh. fixed in  65f67e3

spread/scripts/allocate.sh
lczyk: nope. legacy leftover. cleaned up in 65f67e3

.github/workflows/spread-tests.yml
lczyk: good point. fixed now

spread/scripts/wait-for-ssh.bats
lczyk: good catch. c501746

spread.md
lczyk: fixed in 3d42d65

spread.yaml
lczyk: 👍 moved to top of `systems:` in 5687a5e

spread.yaml
lczyk: we need to source to get `FATAL` and `ADDRESS`. else they don't get passed into the script

spread.yaml
lczyk: nope. the timers are sequential and each script has one kill timer. `allocate/discard`s read only the top-level. `tasks/prepare/restore` walk up the hierarchy and take the first one.

.github/workflows/spread-tests.yml
```diff
@@ -5,67 +5,189 @@ on:
...
-          # We need write access to /dev/kvm for better performance.
-          # Since this is an ephemeral CI/CD environment, we can grant everyone access to /dev/kvm.
           sudo chmod -v 666 /dev/kvm || true
 
-      - name: Cache pristine virtual machine images
-        uses: actions/cache@v6
+      - name: Restore cached vm image
+        id: restore-image
+        if: ${{ !inputs.force_rebuild }}
+        uses: actions/cache/restore@v6
```
dariuszd21: We haven't forced it here yet, but probably we should pin those actions.

Leaving it here, so we can create a follow up

.github/workflows/scripts-tests.yml
```diff
@@ -45,7 +49,9 @@ jobs:
       # not shellcheck-able (bats syntax), so only the plain scripts run.
       - name: Run shellcheck
         run: |
-          shellcheck \
+          shellcheck -x \
```
dariuszd21: What `-x` does ? I cannot find it quickly in [docs](https://github.com/koalaman/shellcheck/wiki/GitHub-Actions#basic-usage)

spread/scripts/wait-for-ssh.sh
```diff
@@ -1,30 +1,41 @@
...
   local interval="${4:-5}"
-  local i
-  for ((i = 0; i < attempts; i++)); do
-    if _ssh_banner_ready "$host" "$port"; then
+  local banner_timeout="${5:-10}"
+  local deadline=$((SECONDS + timeout))
+  local remaining
+  while ((SECONDS < deadline)); do
+    remaining=$((deadline - SECONDS))
+    if _ssh_banner_ready "$host" "$port" "$((remaining < banner_timeout ? remaining : banner_timeout))"; then
```
dariuszd21: If deadline has passed, `_ssh_banner_ready` would get a negative value

spread/scripts/allocate.sh
```diff
@@ -0,0 +1,72 @@
...
+  # image-garden hands out the address once the port-forward exists, which under
+  # emulation can be well before sshd actually answers. Poll for the banner and
+  # hand over as soon as the daemon responds, with a hard 10-minute upper bound
+  # (checked every 5s).
+  # shellcheck source=spread/scripts/wait-for-ssh.sh
+  source spread/scripts/wait-for-ssh.sh
+  if wait_for_ssh "${OUT%:*}" "${OUT##*:}" 600 5; then
+    echo "ssh is answering on $OUT"
+  else
+    FATAL "ssh did not become ready on $OUT"
```
dariuszd21: Is it possible that if this fails, then allocated VM leaks ?

spread/scripts/prepare.sh
```diff
@@ -20,6 +20,10 @@ else
     sudo snap wait system seed.loaded
     sudo snap refresh --hold || true
     sudo snap abort --last=auto-refresh 2>/dev/null || true
+    # The seeded snapd predates fchmodat2 in its seccomp template, which
+    # breaks `docker cp`. A named refresh goes through the hold, and snapd
+    # alone does not reboot the machine.
+    sudo snap refresh snapd
```
dariuszd21: Maybe this is a good idea to do it before `snap refresh` is put on hold (I don't know if it overrides the `--hold` preference or not)?

spread/scripts/wait-for-ssh.bats
```diff
@@ -25,10 +26,14 @@ s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
...
+held = []
 while True:
     c, _ = s.accept()
-    c.sendall(sys.argv[1].encode())
-    c.close()
+    if sys.argv[1]:
+        c.sendall(sys.argv[1].encode())
+        c.close()
+    else:
+        held.append(c)
```
dariuszd21: nit: This is a short-living process, so it's not a big deal, but this socket is never closed

.github/scripts/spread-matrix/spread-matrix
```diff
@@ -0,0 +1,53 @@
...
+matchable="$(paste <(printf '%s\n' "$jobs") <(sed -E 's/-core-([0-9]+)\./-core-\1.04./' <<<"$jobs"))"
+
+# shellcheck disable=SC2086 # word splitting of $FILTER is the point
+for term in $FILTER; do
```
dariuszd21: This can trigger bash filename expansion, so a filter such as `.*` expands into matching filenames in the current directory instead of being passed to `grep` as the intended regular expression.

spread/scripts/prepare-firmware.sh
```diff
@@ -0,0 +1,58 @@
...
+    flash_size=32M # the riscv64 vm requires 32 MiB flash images
+    mkdir -p .image-garden
+    for part in code vars; do
+      src="/snap/image-garden/current/components/qemu-riscv64/share/qemu/edk2-riscv-${part}.fd"
+      dst=".image-garden/efi-${part}.riscv64.img"
+      if [ -f "$src" ] && [ ! -f "$dst" ]; then
+        echo "Pre-building $dst (snap-safe stand-in for image-garden's truncate)"
+        # Plain cp: the fresh mtime makes make consider the target up to date.
+        cp "$src" "$dst"
+        qemu-img resize -f raw "$dst" "$flash_size"
```
dariuszd21: If `qemu-img resize` fails, subsequent calls will use incorrectly sized `$dst`.
It might make sense to remove `$dst` in case of resize failure

dariuszd21 (changes requested):
Thanks for the works on that!

Only few minor correctness concerns left, before it can be merged!

.github/workflows/spread-tests.yml
lczyk: #401

spread/scripts/wait-for-ssh.sh
lczyk: 55f0484

spread/scripts/allocate.sh
lczyk: it is. addressed in 8bda58f

.github/workflows/scripts-tests.yml
lczyk:
```
$ shellcheck --help | grep -- -x
  -x                  --external-sources         Allow 'source' outside of FILES
```

e.g. https://github.com/canonical/docker-snap/blob/3d6fcf057ba0f82a4d025cad75f393fc186825d0/spread/scripts/allocate.sh#L44-L45

spread/scripts/prepare.sh
lczyk: `sudo snap refresh --hold` with no snap name only holds auto refreshes, so this is fine. this is done this way on purpose:

https://github.com/canonical/docker-snap/blob/3d6fcf057ba0f82a4d025cad75f393fc186825d0/spread/scripts/prepare.sh#L20-L26

wait for first boot, stop any new auto refreshed, abort any that may have started and finally selectively refresh only snapd.

the problem statement is that is we dont hold the auto-refresh on core, it might, depending on the image, be refreshing to update kernel or base and both of those would restart the image.

.github/scripts/spread-matrix/spread-matrix
lczyk: noice. good catch! 476ec04

spread/scripts/prepare-firmware.sh
lczyk: 3d6fcf0

spread/scripts/wait-for-ssh.bats
lczyk: this is on purpose. [`start_listener()`](https://github.com/canonical/docker-snap/blob/3d6fcf057ba0f82a4d025cad75f393fc186825d0/spread/scripts/wait-for-ssh.bats#L21-L46) has to accept and then stay silent with the conn open (like qemu's port-forward before the guest sshd is up). that's what the behaviour the test needs. if we close the socket here, the client's read gets `eof` and the test always passes. [`teardown()`](https://github.com/canonical/docker-snap/blob/3d6fcf057ba0f82a4d025cad75f393fc186825d0/spread/scripts/wait-for-ssh.bats#L12-L17) cleans up the socket

PR #389, opened 2026-07-23
fix(ci): repair and harden the nvidia graphics tests

lczyk:
the nvidia graphics tests were failing at the "run tests" step: no testflinger creds were passed, so the cli fell back to an interactive OIDC auth flow that can't work in CI. this PR fixes the auth, then hardens the surrounding test scripts, which had accumulated flakiness and a couple of coverage gaps.

i did *not* refactor the testflinger tests into spread but i think it would be 1) easy and 2) a good idea since it unifies two out of too-many testing paradigms we currently have. i dont want to conflict with https://github.com/canonical/docker-snap/pull/388 too much since it introduces a bunch of spread.yaml changes already but if we merge that first, then rebasing this on top + one more refactor should be quite simple.

## changes

**fixes**

- pass testflinger client-id/secret-key to the submit action
- wait for the device to actually reboot (boot-id change) before testing, instead of racing a still-up pre-reboot box
- quote the docker `-v` mount path in the refresh test

**hardening / ci**

- pin the testflinger submit action to `submit/v1.4.8` (was `@main`, which drifts)
- ssh connect-timeout + keepalive so a mid-refresh reboot can't hang the wait
  loops for minutes
- job `timeout-minutes` and least-privilege `permissions:` on the workflow
- smaller self-hosted runner for the test job (it's just a glorified ssh to testflinger)
- tighten the snap-settle poll interval
- drop the obsolete snapd-beta workaround (stable snapd is now two patch versions ahead of the workaround)
- new `testflinger-scripts-tests.yml`: bats + shellcheck on the testflinger scripts

**coverage**

- `smi_test` now asserts nvidia-smi succeeds and prints a real driver banner, instead of `|| true` swallowing failures
- the refresh test asserts the container is healthy (restart count stable), not merely present -- a crash-looping container no longer passes
- bats coverage for the snap-change parser

## validation

green on real hardware across all three distros (ubuntu-24.04, core22, core24): https://github.com/canonical/docker-snap/actions/runs/30020329114


.github/workflows/nvidia-test.yml
```diff
@@ -74,7 +83,9 @@ jobs:
                 mv $TESTFLINGER_DIR/nvidia-job.temp $TESTFLINGER_DIR/nvidia-job.yaml
 
             - name: Run tests
-              uses: canonical/testflinger/.github/actions/submit@main
+              uses: canonical/testflinger/.github/actions/submit@submit/v1.4.8
```
dariuszd21, line 86: It is generally recommended to use immutable references for actions pinning, and the general pattern I've seen and used is the one below: 
```suggestion
              uses: canonical/testflinger/.github/actions/submit@545ed69f17a5e639a2299e325f5fb199ca9f98af  # submit/v1.4.8
```

This "pinning" will not work if someone enables [require actions to be pinned policy](https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/).

.github/workflows/scripts-tests.yml
```diff
@@ -0,0 +1,46 @@
...
+defaults:
+  run:
+    shell: 'bash -Eeuo pipefail -x {0}'
+
+jobs:
+  bats:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
```
dariuszd21, line 25: Same here and in the `shellcheck` part
```suggestion
        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
```

.github/workflows/testflinger/scripts/check-snap-changes.sh
```diff
@@ -5,28 +5,28 @@
 
 echo "Get snap changes"
 
-# list the snap changes on the device and store the output in a temp file
+# list the snap changes on the device and store the output in a temp file;
+# clean it up on every exit path, this script runs in ssh polling loops
 OUTPUT=$(mktemp)
-snap changes > $OUTPUT
+trap 'rm -f "$OUTPUT"' EXIT
```
dariuszd21, line 11: Is there any particular reason to use `-f`?
As I can tell this file should always exist at this point, and the previous implementation wasn't using `-f`

.github/workflows/testflinger/scripts/update-snapd-beta.sh
dariuszd21, line 1: Why is this no longer needed?

I can see `2.55` and `2.62` being still a thing for non-amd64 releases of Ubuntu 22.04 and 24.04 respectively:
https://packages.ubuntu.com/jammy/snapd
https://packages.ubuntu.com/noble/snapd

.github/workflows/testflinger/scripts/refresh-test.sh
```diff
@@ -51,22 +51,39 @@ EOF
...
+  # crash-looping container fails the test instead of silently passing.
+  CONTAINER=$(sudo docker ps -q)
+  RESTARTS_BEFORE=$(sudo docker inspect -f '{{.RestartCount}}' "$CONTAINER")
+  sleep 5
+  RESTARTS_AFTER=$(sudo docker inspect -f '{{.RestartCount}}' "$CONTAINER")
+  if [ "$RESTARTS_AFTER" -gt "$RESTARTS_BEFORE" ]; then
+    echo "Container is crash-looping (restart count $RESTARTS_BEFORE -> $RESTARTS_AFTER); GPU workload is failing"
+    sudo docker logs "$CONTAINER" 2>&1 | tail -n 20
+    exit 1
+  fi
```
dariuszd21, line 81: Is there a particular reason why those tests use `--restart=always`, so you have to check the restart count?

If the container is a one-shot, then it would be enough if it is running at this point.

.github/workflows/scripts-tests.yml
```diff
@@ -0,0 +1,46 @@
...
+
+# Cancel an in-progress run when a newer commit lands on the same ref, so a
+# burst of pushes to a branch or PR only tests the latest one.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: 'bash -Eeuo pipefail -x {0}'
```
dariuszd21, line 18: It also affects checkout and apt steps, is it intended?
Shouldn't it be scoped to only `bats` step or removed entirely?

.github/workflows/testflinger/scripts/test.sh
```diff
@@ -1,22 +1,38 @@
...
 
   case "$ID-$VERSION_ID" in
     ubuntu-24.04)
-      sudo docker run --rm --runtime=nvidia --gpus all --env PATH="${PATH}:/var/lib/snapd/hostfs/usr/bin" ubuntu nvidia-smi || true
+      smi_out=$(sudo docker run --rm --runtime=nvidia --gpus all --env PATH="${PATH}:/var/lib/snapd/hostfs/usr/bin" ubuntu nvidia-smi)
+      assert_driver "$smi_out"
       ;;
     ubuntu-core-22)
-      sudo docker run --rm --runtime nvidia --gpus all ubuntu bash -c "/snap/docker/*/graphics/bin/nvidia-smi" || true
+      smi_out=$(sudo docker run --rm --runtime nvidia --gpus all ubuntu bash -c "/snap/docker/*/graphics/bin/nvidia-smi")
```
dariuszd21, line 24: The other ones use `--runtime=nvidia` syntax (with equal sign instead of space), maybe it's a good idea to keep them aligned if possible

.github/workflows/testflinger/scripts/test.sh
```diff
@@ -1,22 +1,38 @@
...
-      # Run nvidia-smi from the kernel snap
-      LD_LIBRARY_PATH=/var/snap/pc-kernel/common/kernel-gpu-2404/usr/lib/x86_64-linux-gnu/ /var/snap/pc-kernel/common/kernel-gpu-2404/usr/bin/nvidia-smi || true
+      # In a container, via the nvidia-smi the toolkit mounts from the docker
+      # snap's gpu-2404 component.
+      smi_out=$(sudo docker run --rm --runtime=nvidia --gpus all ubuntu bash -c "/snap/docker/*/gpu-2404*/usr/bin/nvidia-smi")
+      assert_driver "$smi_out"
+      # And on the host, from the kernel snap, which validates the kernel GPU
+      # component itself.
+      smi_out=$(LD_LIBRARY_PATH=/var/snap/pc-kernel/common/kernel-gpu-2404/usr/lib/x86_64-linux-gnu/ /var/snap/pc-kernel/common/kernel-gpu-2404/usr/bin/nvidia-smi)
+      assert_driver "$smi_out"
```
dariuszd21, line 35: It was previously checked on the host first, which seems to make more sense (it won't work in a container if host is missing nvidia kernel snap component?).

Is there any reason for putting it after the container check?

.github/workflows/testflinger/scripts/test.sh
```diff
@@ -1,22 +1,38 @@
...
-      sudo docker run --rm --runtime nvidia --gpus all ubuntu bash -c "/snap/docker/*/graphics/bin/nvidia-smi" || true
+      smi_out=$(sudo docker run --rm --runtime nvidia --gpus all ubuntu bash -c "/snap/docker/*/graphics/bin/nvidia-smi")
+      assert_driver "$smi_out"
       ;;
     ubuntu-core-24)
-      # Run nvidia-smi from the kernel snap
-      LD_LIBRARY_PATH=/var/snap/pc-kernel/common/kernel-gpu-2404/usr/lib/x86_64-linux-gnu/ /var/snap/pc-kernel/common/kernel-gpu-2404/usr/bin/nvidia-smi || true
+      # In a container, via the nvidia-smi the toolkit mounts from the docker
+      # snap's gpu-2404 component.
+      smi_out=$(sudo docker run --rm --runtime=nvidia --gpus all ubuntu bash -c "/snap/docker/*/gpu-2404*/usr/bin/nvidia-smi")
```
dariuszd21, line 30: Is there any suffix after `gpu-2404` in this path or this can be simplified?

```suggestion
      smi_out=$(sudo docker run --rm --runtime=nvidia --gpus all ubuntu bash -c "/snap/docker/*/gpu-2404/usr/bin/nvidia-smi")
```

dariuszd21 (commented):
Thanks for the work, looks really nice.

It's my first review here, but in a principle I'd personally prefer a separate PRs for introduction of shellcheck and the fix.
Now it's a bit hard to navigate what was changed to satisfy the linter versus the actual fix.

alesancor1 (commented):
Looks pretty good, thanks!

I don't have any particular comments other than what @dariuszd21 has already mentioned, please address those 👍

.github/workflows/nvidia-test.yml
lczyk: yep, agreed. thanks for noting the policy too, didn't know about it. will have a look :)

.github/workflows/testflinger/scripts/check-snap-changes.sh
lczyk: i tend to write traps like that since i consider them "best effort" and tbh im still not fully sure what happens when one trap raises but there is another one in flight, *in spite* of writing an entire library (well, i say library... it's more of a single book, page even 😅) around traps https://www.github.com/lczyk/defer.sh

tldr;-f because the nonzero exit is unavoidable -- it tells us nothing since we're unwrapping and tearing everything down anyway.

iff file existence is a test-correctness thing i'd assert it explicitly instead of raising in the cleanup ( *warnings* in the cleanup are fine )

.github/workflows/testflinger/scripts/update-snapd-beta.sh
lczyk: workaround was for kernel-components bug in snapd <2.74; stable is now ≥2.74

.github/workflows/testflinger/scripts/refresh-test.sh
lczyk: test is `run-vector.sh` which is an infinite loop, not one-shot. `--restart=always` + test predate this PR

.github/workflows/scripts-tests.yml
lczyk: this mirrors pre-existing `smoke-test.yml` and `spread-tests.yml`

.github/workflows/testflinger/scripts/test.sh
lczyk: https://github.com/canonical/docker-snap/pull/389/commits/eccc531a8a54a1294e5836fc13bcd04b3a47e669

.github/workflows/testflinger/scripts/test.sh
lczyk: nope. better to check before. changed in https://github.com/canonical/docker-snap/pull/389/commits/d290dfbe3f57986765e90fc8075651f7c2b9538f

.github/workflows/scripts-tests.yml
dariuszd21: It might be worth looking at some point, if it is safe and/or expected to pass that to the actions that are not owned by us.

.github/workflows/testflinger/scripts/update-snapd-beta.sh
dariuszd21: So those tests run only on x86_64, if that's the case it's fine.
I just wanted to point out here that other CPU architectures don't have snapd 2.74 :smile:

.github/workflows/testflinger/scripts/check-snap-changes.sh
dariuszd21: For me that's the sign, that the script should be ported to Python :sweat_smile:

dariuszd21 (approved):
Thanks for addressing the review.
LGTM!

PR #390, opened 2026-08-24
chore: update moby to v29.7.2

dariuszd21:
containerd is not updated here, as this version of moby still depends on `containerd` v2.3.3.

`engine` and `docker-cli` parts now use `$CRAFT_PROJECT_VERSION`, so the version doesn't have to be repeated in 3 places.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: docker
-version: '29.6.1'
+version: '29.7.2'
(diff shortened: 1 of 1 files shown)


lczyk (approved):

lczyk:
looks good (i've tested patches, look over the release noted and confirmed all is running fine). merged to get it to edge and full spread / nvidia-graphics testing.

PR #391, opened 2026-08-26
feat: set daemon options via snap config

### LINKED ISSUE #299: content slot for rw access to docker daemon config

lczyk:
let a deployment set specific daemon.json options (e.g. `default-address-pools`) without exposing a read-write content interface onto the daemon config.

this PR exposes some daemon.json options through `snap set docker daemon.<key>=<value>`. only a whitelist of safe options is accepted. options that expose or take over the daemon, or that the snap passes to dockerd as flags are excluded. a key outside the allowlist (a typo or an unsupported option) is rejected. the merged config is validated with `dockerd --validate`.

alternative to #328
closes #299


lib/daemon-config
```diff
@@ -0,0 +1,95 @@
...
+# hook merges into daemon.json. Whitelist safe keys. Keys with a value in the base config
+# (e.g. log-level) are restored to that default when unset.
+#
+# Known dangerous keys are: hosts, tls*, authorization-plugins, runtimes,
+# insecure-registries, registry-mirrors. We should have a good reason for adding
+# them since they could grant network/root access.
+
+USER_DAEMON_KEYS=(
+    default-address-pools
+    bip
```
dariuszd21, line 16: For other ones `v6` support is provided. 
Should `bip6` be added to the whitelist ?

lib/daemon-config
```diff
@@ -0,0 +1,95 @@
...
+    fixed-cidr
+    fixed-cidr-v6
+    default-gateway
+    default-gateway-v6
+    mtu
+    dns
+    dns-search
+    dns-opts
```
dariuszd21, line 23: I think it make sense to follow lexicographical order here, same as in [docs](https://docs.docker.com/reference/cli/dockerd/#daemon-configuration-file)
```suggestion
    dns
    dns-opts
    dns-search
```

lib/daemon-config
```diff
@@ -0,0 +1,95 @@
...
+    default-ulimits
+    log-level
+)
+
+reject_unknown_daemon_config() {
+    local allow doc bad
```
dariuszd21: Why is it called `doc` 😄 ?

lib/daemon-config
```diff
@@ -0,0 +1,95 @@
...
+    bip
+    fixed-cidr
+    fixed-cidr-v6
+    default-gateway
+    default-gateway-v6
+    mtu
+    dns
+    dns-search
+    dns-opts
+    default-ulimits
```
dariuszd21: This one seems a bit tricky as this requires passing the whole json nested object

lib/daemon-config
```diff
@@ -0,0 +1,95 @@
...
+
+# Keys a user may set via `snap set docker daemon.<key>=<value>`, which the configure
+# hook merges into daemon.json. Whitelist safe keys. Keys with a value in the base config
+# (e.g. log-level) are restored to that default when unset.
+#
+# Known dangerous keys are: hosts, tls*, authorization-plugins, runtimes,
+# insecure-registries, registry-mirrors. We should have a good reason for adding
+# them since they could grant network/root access.
+
+USER_DAEMON_KEYS=(
```
dariuszd21, line 15: Can we sort all the keys here ?

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,58 @@
...
+  # successive changes below would otherwise trip systemd's start-limit, so reset it first.
+  snapcfg() {
+    sudo systemctl reset-failed snap.docker.nvidia-container-toolkit.service 2>/dev/null || true
+    sudo snap "$@"
+  }
+
+  # An allowlisted option is merged into daemon.json and takes effect.
+  snapcfg set docker daemon.default-address-pools='[{"base":"10.211.0.0/16","size":24}]'
+  wait_for_docker
+  sudo grep -q '10.211.0.0/16' "$CONFIG"
```
dariuszd21: There is `MATCH` helper in spread for grepping

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+  sudo docker network rm testpool >/dev/null 2>&1 || true
+  sudo docker network create testpool
+  subnet=$(sudo docker network inspect testpool -f '{{range .IPAM.Config}}{{.Subnet}}{{end}}')
+  echo "testpool subnet: $subnet"
+  echo "$subnet" | grep -q '^10.211.' || ERROR "created network did not draw from the configured pool"
```
dariuszd21: MATCH seems to be also a better fit here

README.md
```diff
@@ -108,6 +108,16 @@ Docker should function normally, with the following caveats:
 
 * Specifying the option `--security-opt="no-new-privileges=true"` with the `docker run` command (or the equivalent in docker-compose) will result in a failure of the container to start. This is due to an an underlying external constraint on AppArmor; see [LP#1908448](https://bugs.launchpad.net/snappy/+bug/1908448) for details.
 
+### Configuring daemon options
+
+A subset of `daemon.json` options can be set through snap configuration. This is convenient for unattended and Ubuntu Core deployments, where the values can be provided from a gadget's `defaults`:
```
dariuszd21: I wouldn't add the part about Ubuntu Core. The value here is that is might be always convenient to set it

lib/daemon-config
```diff
@@ -0,0 +1,95 @@
...
+    default-gateway-v6
+    mtu
+    dns
+    dns-search
+    dns-opts
+    default-ulimits
+    log-level
+)
+
+reject_unknown_daemon_config() {
```
dariuszd21: Should we call it unsupported rather than unknown?

README.md
```diff
@@ -108,6 +108,16 @@ Docker should function normally, with the following caveats:
...
+
+```shell
+sudo snap set docker daemon.default-address-pools='[{"base":"10.0.0.0/8","size":24}]'
+```
+
+Only a subset of options is accepted: `default-address-pools`, `bip`, `fixed-cidr`, `fixed-cidr-v6`, `default-gateway`, `default-gateway-v6`, `mtu`, `dns`, `dns-search`, `dns-opts`, `default-ulimits` and `log-level`. Setting any other daemon option this way fails with an error. An option set this way takes precedence over the same option in `daemon.json` and is written back on every configuration change, so change it with `snap set` and remove it with `snap unset` rather than editing `daemon.json`. Options that were never set through snap configuration are left as they are in the file.
```
dariuszd21: It might be worth sorting options here as well

lib/daemon-config
```diff
@@ -0,0 +1,128 @@
...
+    prev="$(cat "$managed" 2>/dev/null || true)"
+    jq -e 'type == "array" and all(type == "string")' <<<"$prev" >/dev/null 2>&1 || prev='[]'
+
+    # Defaults for whitelisted keys, so unsetting a key that
+    # has a base default restores that default instead of dockerd's built-in.
+    base="$(jq -c --argjson allow "$allow" \
+        'with_entries(select(.key as $k | $allow | index($k)))' "${SNAP}/config/daemon.json" 2>/dev/null || true)"
+    jq -e 'type == "object"' <<<"$base" >/dev/null 2>&1 || base='{}'
+
+
```
dariuszd21, line 106: (via ragrev) Any particular reason to have 2 blank lines here?

lib/daemon-config
```diff
@@ -0,0 +1,128 @@
...
+            echo "error: existing ${cfg} is invalid: ${cfg_err}" >&2
+        else
+            echo "error: invalid daemon option value: ${err}" >&2
+        fi
+        return 1
+    fi
+
+    # Only apply when the config actually changed
+    if jq -S . "$cfg" 2>/dev/null | cmp -s - <(jq -S . "$tmp" 2>/dev/null); then
+        rm -f "$tmp"
```
dariuszd21: Should it return here ?
The 127 line runs no matter the outcome of this `if`

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,89 @@
...
+  wait_for_docker
+  sudo grep -q '"snap.test"' "$CONFIG" || ERROR "hand-set option was dropped by an unrelated snap set"
+  sudo grep -qE '"log-level"[[:space:]]*:[[:space:]]*"debug"' "$CONFIG" || ERROR "hand-set log-level was reset by an unrelated snap set"
+  sudo grep -qE '"mtu"[[:space:]]*:[[:space:]]*1400' "$CONFIG" || ERROR "mtu was not applied"
+
+  # Once a key has been set through snap config it is managed: unset restores the shipped
+  # default, not the hand-set value.
+  snapcfg set docker daemon.log-level=info
+  wait_for_docker
+  snapcfg unset docker daemon.log-level
```
dariuszd21, line 79: It might be not a big thing, but this operation resets the value to the factory default and forget about user setting it to `debug` in the first place.


If `configure` hook is called during refresh this may end up removing user-set configuration when package is updated to the newer revision

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,89 @@
...
+  snapcfg unset docker daemon.default-address-pools
+  wait_for_docker
+  ! sudo grep -q '10.211.0.0/16' "$CONFIG" || ERROR "option was not removed on unset"
+
+restore: |
+  source "$SCRIPTS_PATH/common.sh"
+
+  sudo systemctl reset-failed snap.docker.nvidia-container-toolkit.service 2>/dev/null || true
+  sudo snap unset docker daemon.default-address-pools daemon.hosts daemon.mtu daemon.log-level 2>/dev/null || true
+  sudo snap run --shell docker.docker -c "jq 'del(.\"dns-search\")' \"\$SNAP_DATA/config/daemon.json\" > \"\$SNAP_DATA/config/daemon.json.new\" && mv \"\$SNAP_DATA/config/daemon.json.new\" \"\$SNAP_DATA/config/daemon.json\""
```
dariuszd21: It might simplify things a bit if there is a copy of the configuration made during `prepare:` step and here it is restored

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,89 @@
...
+  echo "testpool subnet: $subnet"
+  echo "$subnet" | grep -q '^10.211.' || ERROR "created network did not draw from the configured pool"
+  sudo docker network rm testpool
+
+  # A dangerous / unknown option is rejected outright: the configure hook fails the
+  # `snap set`, so nothing is stored, nothing reaches daemon.json, and no port is opened.
+  ! snapcfg set docker daemon.hosts='["tcp://0.0.0.0:2375"]' || ERROR "disallowed 'hosts' option was accepted"
+  wait_for_docker
+  ! sudo grep -q 'tcp://0.0.0.0:2375' "$CONFIG" || ERROR "'hosts' leaked into daemon.json"
+  ! sudo ss -tlnH 'sport = :2375' | grep -q . || ERROR "daemon exposed a TCP port via a disallowed option"
```
dariuszd21: This will fail if anything listens on 2375 port, can be unrelated to the configuration change all together

dariuszd21 (changes requested):
Thank you for the work here.
I believe this is the way it should look like - vetted set of options exposed via the snap configuration primitives.
My main concern is around `unset`, it might have an unexpected side effect of removing configuration for users not using snap set at all.

lib/daemon-config
lczyk: this if decides whether we have a new config on hand and `mv`s it if so. we still have to update the managed list in 127 (unset that changes nothing still has to drop the key from the managed list). I've added a clarifying note.

lib/daemon-config
lczyk: emmm... no reason 😅 . renamed now.

spread/main/daemon_config/task.yaml
lczyk: the configure bit is fixed in 8935c90. regarding reset to factory and not one which user manually set in deamon.json; yeah, its a bit of an edge case. i'm hesitant to add more things to this PR to do more config juggling, but i think it *could* be implemented. i could also be left as a future improvement.. ? ( unsure how big if a deal this is )

for clarity, the failing case is e.g.:

0. default value of e.g. `default-ulimits` is `foo`
1. user manually sets it in deamon.json. `default-ulimits=bar`
2. `sudo snap set docker deamon.default-ulimits=baz`
3. `sudo snap unset docker deamon.default-ulimits`
4. `default-ulimits` is now `foo` **not** `bar` aka the default value, not the one manually set.

lczyk: for now i've added a note in ed06a14.

spread/main/daemon_config/task.yaml
lczyk: ed06a14

spread/main/daemon_config/task.yaml
lczyk: good spot. cd1a2f3

lib/daemon-config
lczyk: yeah, but it *does* get validated (by `dockerd --validate`) and the example in README shows the shape of it. i'm not really seeing a way to make this simpler, bar writing our own parser and allowing inputs like `... daemon.default-ulimits='nofile=64000:64000'`, and i do not this that's really needed or a good idea. lmk if i'm missing something though! :)

lib/daemon-config
lczyk: yep, worth it. 67f763a.

spread/main/daemon_config/task.yaml
lczyk: ok, i came up with a better design for this, i think. 3 layers: base (B) which is what we ship in snap and is immutable. hand-edits (H) applied on top by whoever whenever. finally options (O) from `snap config`. result (R) is `R = B < H < O`. ok, one more ting actually, the hand-edit could delete a key from B too, so we need tombstones (T) to mark which keys are now gone, so `R = (B - T) < H < O`.

lczyk: it does make the loc larger but at this point i've poured over both this and previous version and, honestly, i think this one is also easier to understand

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,142 @@
+summary: Set daemon.json options via snap configuration
+
+environment:
+  CONFIG: /var/snap/docker/current/config/daemon.json
+  STATE: /var/snap/docker/current/daemon-config-state
+  BACKUP: daemon.json.bak
```
dariuszd21, line 6: You might want to also use the absolute path here too

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,142 @@
...
+
+prepare: |
+  sudo cp "$CONFIG" "$BACKUP"
```
dariuszd21, line 9: Should it have `rm -f "$BACKUP"` beforehand?

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,142 @@
...
+
+execute: |
+
```
dariuszd21, line 12:
```suggestion
```

lib/daemon-config
```diff
@@ -0,0 +1,245 @@
...
+def slice($keys): with_entries(select(.key as $k | $keys | index($k)));
+def unslice($keys): with_entries(select(.key as $k | $keys | index($k) | not));
+'
+
+# Merge whitelisted daemon options from snap config into daemon.json.
+#
+# The whitelisted slice of daemon.json is composed from three layers, later wins:
+#   B: the defaults shipped in $SNAP/config/daemon.json
+#   H: values set by hand in daemon.json
+#   O: snap config (`snapctl get daemon`)
```
dariuszd21, line 84:
```suggestion
#   O: snap config (`snapctl get -d daemon`)
```

lib/daemon-config
```diff
@@ -0,0 +1,245 @@
...
+    # committed unconditionally.
+    mv "${statef}.new" "$statef"
+}
+
+# Copy daemon.json and the merge state aside, for restore_daemon_config.
+snapshot_daemon_config() {
+    rm -rf "$DAEMON_CONFIG_ROLLBACK" && mkdir -p "$DAEMON_CONFIG_ROLLBACK"
+    cp "$DAEMON_CONFIG_FILE" "${DAEMON_CONFIG_ROLLBACK}/daemon.json"
+    [ ! -f "$DAEMON_CONFIG_STATE" ] || cp "$DAEMON_CONFIG_STATE" "${DAEMON_CONFIG_ROLLBACK}/state"
+}
```
dariuszd21, line 227: [ragrev]
 `snapshot_daemon_config` unconditionally `cp`s `$DAEMON_CONFIG_FILE` without checking it exists, unlike the guarded copy for `$DAEMON_CONFIG_STATE` right below it. If `daemon.json` is ever missing (e.g., first boot before the file is created, or some corrupted install state), this `cp` fails under `set -e` in the configure hook and aborts the whole hook rather than degrading gracefully.

snap/snapcraft.yaml
```diff
@@ -126,6 +126,7 @@ parts:
       - mount
       - nftables
     organize:
+      lib/daemon-config: usr/share/docker/daemon-config
```
dariuszd21, line 129: That seems a bit dangerous to place it here: it makes an impression that this file is was shipped by docker deb package
Why do you need to reorganize it?

lib/daemon-config
```diff
@@ -0,0 +1,245 @@
...
+
+# Put the snapshot back. Returns 0 iff daemon.json actually changed, so the caller
```
dariuszd21:
```suggestion
# Put the snapshot back. Returns 0 if daemon.json actually changed, so the caller
```

lib/daemon-config
```diff
@@ -0,0 +1,245 @@
...
+           state: {result: $R, hand: $H, tombstone: ($T | sort)},
+           changed: ($R != $C)}' \
+        --argjson allow "$allow" --argjson B "$base" --argjson O "$overlay" \
+        --argjson state "$state" "$cfg" 2>/dev/null)"; then
+        echo "error: failed to merge daemon options into ${cfg}" >&2
+        return 1
+    fi
+
+    tmp="$(mktemp "${cfg}.XXXXXX")"
+    chmod 644 "$tmp" # mktemp creates it 0600
```
dariuszd21, line 189: Is the permission being 0600 a problem?

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,142 @@
...
+restore: |
+  source "$SCRIPTS_PATH/common.sh"
+
+  sudo systemctl reset-failed snap.docker.nvidia-container-toolkit.service 2>/dev/null || true
+  # Unset before restoring the file: each unset runs the configure hook, which rewrites
+  # daemon.json and the merge state.
+  sudo snap unset docker daemon.default-address-pools daemon.hosts daemon.mtu daemon.log-level 2>/dev/null || true
+  sudo cp "$BACKUP" "$CONFIG"
+  sudo rm -f /var/snap/docker/current/daemon-config-state /var/snap/docker/current/daemon-config-keys
```
dariuszd21, line 138:
```suggestion
  sudo rm -f "${STATE}" /var/snap/docker/current/daemon-config-keys
```

dariuszd21: What's the second file here, cannot file `daemon-config-keys` anywhere else

lib/daemon-config
```diff
@@ -0,0 +1,245 @@
+#!/bin/bash
+# Helpers for the configure hook to manage user-set daemon.json options via
+# `snap set docker daemon.<key>=<value>`.
+
+# Keys a user may set via `snap set docker daemon.<key>=<value>`, which the configure
+# hook merges into daemon.json. Whitelist safe keys.
+#
+# Known dangerous keys are: hosts, tls*, authorization-plugins, runtimes,
+# insecure-registries, registry-mirrors. We should have a good reason for adding
+# them since they could grant network/root access.
```
dariuszd21, line 10: This is viable information, but i can be easily missed during the review.

If someone will add any of those in the USER_DAEMON_KEYS, you won't see this message in a diff, so it's very easy to miss.

Maybe the idea here could be a follow up PR with explicitly rejecting those in the script, so whenever someone decides to change it, the diff will be suspicious

lib/daemon-config
```diff
@@ -0,0 +1,245 @@
...
+#   B: the defaults shipped in $SNAP/config/daemon.json
+#   H: values set by hand in daemon.json
+#   O: snap config (`snapctl get daemon`)
+# and four more names take part:
+#   T: tombstones, hand-deleted B keys that composing must not resurrect
+#   C: the current file, as it is now
+#   P: what the previous run wrote
+#   R: the composed result; stored in the state, it is the next run's P
+# H and T are reconstructed each run by diffing C against P. P, H and T live in
+# $DAEMON_CONFIG_STATE as {result, hand, tombstone}. Hand-writing a key back to its
```
dariuszd21, line 91: Something is not right here. 
According to this description `P` live in `$DAEMON_CONFIG_STATE` as `{result}`

farshidtz (commented):
Configuration via snap options is surely superior to https://github.com/canonical/docker-snap/pull/328. This is a decent implementation, given only a subset of keys are exposed and that everything gets validation before being applied.

I suggest adding smoke tests as well to test the changes on PRs, as well as refresh testing to make sure there configurations persist across refreshes (upgrade and downgrade).

Keep in mind that dropping support for a key in the future would be a backward incompatible change, unless gracefully handled via a pre/post-refresh hook. For example, if dockerd stops accepting `bip` and the snap is forced to drops it too, but the user has set a value for it via the corresponding snap option, the snap would fail to refresh. To prevent that, a logic should be added to the refresh hook to unset/replace the configuration.

lib/daemon-config
```diff
@@ -0,0 +1,245 @@
...
+DAEMON_CONFIG_STATE="${SNAP_DATA}/daemon-config-state"
+DAEMON_CONFIG_ROLLBACK="${SNAP_DATA}/.configure-rollback"
+
+reject_unsupported_daemon_config() {
+    local snap_cfg bad
+    snap_cfg="$(snapctl get -d daemon 2>/dev/null || true)"
+    [ -n "$snap_cfg" ] || return 0
+    # `snap set docker daemon=foo` arrives as a non-object; reject it with a
+    # message instead of letting jq abort the hook silently under set -e.
+    if ! jq -e '(.daemon // {}) | type == "object"' <<<"$snap_cfg" >/dev/null 2>&1; then
```
dariuszd21, line 45: I'm not an expert here, but according to my GPT colleague it can be bypassed:

Medium: **daemon=false** bypasses the non-object check and removes active overrides.
[daemon-config, line 45](https://github.com/canonical/docker-snap/blob/97c682c636b71b00a21d8c5b9d66ec27a543a1ac/lib/daemon-config#L45) uses (.daemon // {}). In jq, // substitutes its fallback for both null and false. Consequently, snap set docker daemon=false is treated as an empty object, rather than rejected.

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,142 @@
+summary: Set daemon.json options via snap configuration
```
dariuszd21, line 1: Shouldn't it be a part of the smoke tests suite ?

snap/hooks/configure
```diff
@@ -16,6 +18,26 @@ fi
...
+snapshot_daemon_config
+on_exit() {
+    rc=$?
+    if [ "$rc" -ne 0 ] && restore_daemon_config; then
+        # the failed restart may have left services stopped or on the new config
+        snapctl restart "${SNAP_NAME}" 2>/dev/null || true
+    fi
+    rm -rf "$DAEMON_CONFIG_ROLLBACK"
+    exit "$rc"
+}
```
dariuszd21, line 34: I believe `snap restart` here is a wrong recovery step.
The configure step fails, you restore the old configuration and unconditionally request the restart of all services (that were not able to pick up the new configuration yet).

lib/daemon-config
```diff
@@ -0,0 +1,245 @@
...
+    # dockerd needs $SNAP/bin on PATH to locate docker-proxy during validation.
+    if err="$(PATH="${SNAP}/bin:${SNAP}/usr/bin:${PATH}" "${SNAP}/bin/dockerd" \
+        --validate --config-file="$1" 2>&1)"; then
+        return 0
+    fi
+    echo "${err#unable to configure the Docker daemon with file *: }"
+    return 1
+}
+
+# jq helpers shared by the programs below, so the whitelist filter cannot drift
```
dariuszd21, line 71: nit: filter cannot drift sounds like the result of calling it is always the same.
I believe the intent was to emphasize that filtering logic is common

lib/daemon-config
```diff
@@ -0,0 +1,245 @@
...
+# between call sites. unslice is slice's complement over any key array.
+# shellcheck disable=SC2016
+DAEMON_CONFIG_JQ='
+def slice($keys): with_entries(select(.key as $k | $keys | index($k)));
+def unslice($keys): with_entries(select(.key as $k | $keys | index($k) | not));
```
dariuszd21, line 76: nit: the `unslice` name for me doesn't convey an intent well.
What about `include_keys` / `exclude_keys` or `keep_keys` / `drop_keys`?

README.md
```diff
@@ -108,6 +108,16 @@ Docker should function normally, with the following caveats:
...
 
+### Configuring daemon options
+
+A subset of `daemon.json` options can be set through snap configuration, without editing the file:
+
+```shell
+sudo snap set docker daemon.default-address-pools='[{"base":"10.0.0.0/8","size":24}]'
+```
+
+Only a subset of options is accepted: `bip`, `bip6`, `default-address-pools`, `default-gateway`, `default-gateway-v6`, `default-ulimits`, `dns`, `dns-opts`, `dns-search`, `fixed-cidr`, `fixed-cidr-v6`, `ipv6`, `log-level` and `mtu`. Setting any other daemon option this way fails with an error. An option set this way takes precedence over the same option in `daemon.json`; options set by hand in the file are preserved, and `snap unset` puts back what the option was before: the hand-set value if there was one, otherwise the value shipped with the snap, otherwise the option is removed. An option you deleted from the file by hand stays deleted until you set it again, and writing the shipped value into the file by hand makes the option follow the snap's default again.
```
dariuszd21, line 119: I'm a bit inclined to restore the previously user-set configuration and especially making it a strong statement/commitment in the documentation.

I've asked the model to come up with the scenario to break it and it's pretty straightforward:
```
Shipped default:                  error
User manually configures:         debug
User snap-sets:                   error
User manually writes:             error
User unsets the snap option:      debug comes back
```

dariuszd21 (changes requested):
First pass done, sorry it took so long but this PR requires a big context window 😅 

I think restoring manually edited configuration after `set` / `unset` unnecessarily complicates things and doesn't quite deliver on the promise. We only see the file when the configure hook runs and have to infer what the user changed, which makes the restoration code bugprone.

Could we just treat snap set as taking ownership of that key, and have snap unset restore the shipped default / remove it ?

lib/daemon-config
lczyk: agreed. made an issue for a followup: https://github.com/canonical/docker-snap/issues/399

lib/daemon-config
lczyk: fixed in d060dfa and covered by test 👍

lib/daemon-config
lczyk: without it, the first snap set would quietly turn a 644 daemon.json into a 600 one (since its writing through tmp). ive updated this in 5c9aaf0 though, and now the mode is preserved.

lczyk:
i've checked this PR against the NVIDIA tests [here](https://github.com/canonical/docker-snap/actions/runs/35356004593) and the full spread suite for amd64 [here](https://github.com/canonical/docker-snap/actions/runs/35377577091), both green.

.github/workflows/smoke-test.yml
```diff
@@ -360,3 +360,28 @@ jobs:
...
+          CONFIG=/var/snap/docker/current/config/daemon.json
+          # each configure run bounces the nvidia oneshot; keep it clear of systemd's start limit
+          snapcfg() {
+            sudo systemctl reset-failed snap.docker.nvidia-container-toolkit.service || :
+            sudo snap "$@"
+          }
+
+          snapcfg set docker daemon.mtu=1400
+          sudo jq -e '.mtu == 1400' "$CONFIG"
+          timeout 60 sh -c 'until sudo docker info >/dev/null 2>&1; do sleep 1; done'
```
dariuszd21, line 376: Can you somehow check it through the CLI as well ?
Then it would tell you that the new configuration was actually loaded

tests/daemon-config.bats
```diff
@@ -0,0 +1,299 @@
+#!/usr/bin/env bats
+# Unit tests for lib/daemon-config: the lib is sourced into the test shell with the
+# snap environment pointed at a temp tree (see daemon-config-fixture.bash).
+
+load daemon-config-fixture
+
+setup() {
+  _setup_snap_tree
+  # shellcheck disable=SC1091
```
dariuszd21: In other places you don't have to disable it, but you just provide a path for shell checker to fallback to, any reason why is it not done here that way?

.github/workflows/scripts-tests.yml
```diff
@@ -41,6 +44,8 @@ jobs:
 
       # shellcheck is preinstalled on the runner image. The .bats files are
       # not shellcheck-able (bats syntax), so only the plain scripts run.
+      # SC1091 is off because the hook sources files by their in-snap path.
```
dariuszd21: It's disabled both here and in `tests/daemon-config.bats`.
I'd suggest and prefer disabling it just locally where needed to avoid side-effects of things not being checked were they should be

snap/hooks/configure
```diff
@@ -15,6 +17,31 @@ fi
...
+# a service restart error) cannot leave the file claiming a value that snap get
+# no longer shows.
+snapshot_daemon_config
+on_exit() {
+    rc=$?
+    if [ "$rc" -ne 0 ] && restore_daemon_config && $SVC_STOPPED; then
+        # the restart failed part-way; bring the services back up on the restored config
+        snapctl stop "${SNAP_NAME}" 2>/dev/null || true
+        snapctl start "${SNAP_NAME}" 2>/dev/null || true
+    fi
```
dariuszd21: Please double-check this behavior with `stop` and `start`. If I remember correctly, `snapctl stop/start` from the configure hook isn't synchronous, but rather a request to snapd to do it after the hook finishes. `SVC_STOPPED` doesn't mean the services have actually stopped, and requesting another stop/start here seems unnecessary. Restoring the files should be enough for file recovery.

lib/daemon-config
```diff
@@ -0,0 +1,180 @@
...
+# Copy daemon.json and the managed-key list aside, for restore_daemon_config.
+snapshot_daemon_config() {
+    rm -rf "$DAEMON_CONFIG_ROLLBACK" && mkdir -p "$DAEMON_CONFIG_ROLLBACK"
+    cp "$DAEMON_CONFIG_FILE" "${DAEMON_CONFIG_ROLLBACK}/daemon.json"
+    [ ! -f "$DAEMON_CONFIG_KEYS" ] || cp "$DAEMON_CONFIG_KEYS" "${DAEMON_CONFIG_ROLLBACK}/daemon-config-keys"
+}
+
+# Put the snapshot back. Returns 0 iff daemon.json actually changed, so the caller
+# knows a service restart is warranted.
+restore_daemon_config() {
```
dariuszd21, line 161: There is an issue that might lead to partial rollback here. 

If the hook fails after the NVIDIA cleanup, it can restore the NVIDIA runtime entry while the NVIDIA configuration files (e.g. CDI) remain deleted, leaving the configuration inconsistent.

lib/daemon-config
```diff
@@ -0,0 +1,180 @@
...
+    local allow snap_cfg managed base overlay merged tmp mode err cfg_err
+    allow="$(daemon_config_whitelist)"
+    snap_cfg="$(snapctl get -d daemon)" || return 1
+    [ -n "$snap_cfg" ] || snap_cfg='{}'
+
+    # -s so a file holding more than one document is discarded, not carried on as two
+    managed="$(jq -cse 'select(length == 1) | .[0] | arrays | select(all(type == "string"))' \
+        "$mng" 2>/dev/null || echo '[]')"
+    base="$(jq -ce "$DAEMON_CONFIG_JQ"'keep_keys($allow)' --argjson allow "$allow" \
+        "${DAEMON_CONFIG_BASE}" 2>/dev/null || echo '{}')"
```
dariuszd21: Hmm should we treat it as an error instead ?
It silently accepts that configuration shipped by the package is broken 😄

lib/daemon-config
```diff
@@ -0,0 +1,180 @@
...
+    if ! overlay="$(jq -ce "$DAEMON_CONFIG_JQ"'(.daemon // {}) | keep_keys($allow)' \
+        --argjson allow "$allow" <<<"$snap_cfg" 2>/dev/null)"; then
+        echo "config: could not read daemon options; leaving daemon.json unchanged"
+        return 0
+    fi
```
dariuszd21: Why this is not an error ?

spread/main/daemon_config/task.yaml
```diff
@@ -0,0 +1,112 @@
...
+  sudo docker network create testpool
+  subnet=$(sudo docker network inspect testpool -f '{{range .IPAM.Config}}{{.Subnet}}{{end}}')
+  echo "testpool subnet: $subnet"
+  echo "$subnet" | MATCH '^10\.211\.' || ERROR "created network did not draw from the configured pool"
+  sudo docker network rm testpool
+
+  # A dangerous / unknown option is rejected outright: the configure hook fails the
+  # `snap set`, so nothing is stored, nothing reaches daemon.json, and no port is opened.
+  # The port check names dockerd, so an unrelated listener on 2375 does not fail the task.
+  ! snapcfg set docker daemon.hosts='["tcp://0.0.0.0:2375"]' || ERROR "disallowed 'hosts' option was accepted"
```
dariuszd21, line 37: Does it make sense to check if error messages are correct in places where `set` fails?

dariuszd21 (changes requested):
I like this approach, it's clean and seem to cover the cases that we care about.
Just few small things to fix and it's ready to be shipped

spread/main/daemon_config/task.yaml
lczyk: already checked in bats tests. spread tests are a bit more end-2-end-y, and so i have a vague sense that maybe we should not try to nail down the exact wording but the behaviour here..

tbh i dont have a strong pref here. if you say i'll add it to spread too.

lib/daemon-config
lczyk: i think lets leave this for later improvement, since this is more of the nvidia hook issue. wdyt? tracked here: https://github.com/canonical/docker-snap/issues/400

snap/hooks/configure
lczyk: > If I remember correctly, snapctl stop/start from the configure hook isn't synchronous, but rather a request to snapd to do it after the hook finishes

"you're absolutely right!" ;P that is, if i understand the snapd source correctly. i believe that the queue [here](https://github.com/canonical/snapd/blob/1f194f650b92f2fbef135231303fa1e711c3d4af/overlord/hookstate/ctlcmd/helpers.go#L487-L494) is exactly what you described, no?

i've assumed you're right and tentatively simplified this to just the file restore in a373217.

lczyk: dug out [this](https://forum.snapcraft.io/t/restarting-services-from-configure-hook-race-condition/2513/6) snapcraft forum post which talks about the same thing, but also goes a bit  beyond my current snap knowledge. linking it here for reference

lib/daemon-config
lczyk: 7c3f7c2

lib/daemon-config
lczyk: no good reason. made it into an error in 7c3f7c2. this should be unreachable from the hook anyway since `reject_unsupported_daemon_config` already fails on the same input.

.github/workflows/scripts-tests.yml
lczyk: yep. agreed. 8dd13b3

tests/daemon-config.bats
lczyk: no reason, switched to source=lib/daemon-config in 8dd13b3

.github/workflows/smoke-test.yml
lczyk: yep. c6354be

lczyk: also found 2 more places: https://github.com/canonical/docker-snap/pull/391/commits/5626b140a462246c267379e21f357b59765ed1f6

lib/daemon-config
dariuszd21: Agree, we shouldn't grow this.
It's also a condition that would need many things to fail in the meantime

dariuszd21 (approved):
Thanks for the work on this!

I think it's ready to give it a go 🎉

PR #392, opened 2026-08-26
feat(ci): nightly smoke test for snapd

lczyk:
this PR adds a nightly smoke test run which runs the smoke tests against stable and beta channels of snapd. `push` and `pull_request` triggers default to just latest/stable.

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -88,9 +88,9 @@ jobs:
         # channel it asked for, defaulting to what users actually run.
         snapd_channel: >-
           ${{ fromJSON(
-              github.event_name == 'schedule'
-              && '["latest/stable","latest/beta"]'
-              || format('["{0}"]', inputs.snapd_channel || 'latest/stable')
+          github.event_name == 'schedule'
+          && '["latest/stable","latest/beta"]'
+          || format('["{0}"]', inputs.snapd_channel || 'latest/stable')
           ) }}
         runner:
           - arch: amd64
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -24,11 +24,27 @@ on:
       - '.github/workflows/nvidia-test.yml'
   # Manual trigger
   workflow_dispatch:
+    inputs:
+      snapd_channel:
+        description: "`snapd` snap channel to test against"
+        required: false
+        type: choice
+        options:
+          - latest/stable
+          - latest/candidate
+          - latest/beta
+          - latest/edge
+        default: latest/stable
+  # Nightly test of latest/stable and latest/beta
+  schedule:
+    - cron: '17 5 * * *'
 
 # Cancel an in-progress run when a newer commit lands on the same ref, so a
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -24,11 +24,27 @@ on:
...
+      snapd_channel:
+        description: "`snapd` snap channel to test against"
+        required: false
+        type: choice
+        options:
+          - latest/stable
+          - latest/candidate
+          - latest/beta
+          - latest/edge
+        default: latest/stable
```
dariuszd21, line 37: This will block running those tests on branch channels (the ones that nvidia workflows upload to)

.github/workflows/smoke-test.yml
```diff
@@ -98,12 +122,29 @@ jobs:
...
+          SNAPD_CHANNEL: ${{ matrix.snapd_channel }}
         run: |
-          sudo snap install snapd
+          if snap list snapd >/dev/null 2>&1; then
+            sudo snap refresh snapd --channel="$SNAPD_CHANNEL"
+          else
+            sudo snap install snapd --channel="$SNAPD_CHANNEL"
+          fi
+
+          sudo snap watch --last=refresh || true
```
dariuszd21, line 136: If you want to watch, then maybe `refresh` should use `--no-wait` ?
If I remember correctly refresh is synchronous by default

.github/workflows/smoke-test.yml
```diff
@@ -24,11 +24,27 @@ on:
...
+        type: choice
+        options:
+          - latest/stable
+          - latest/candidate
+          - latest/beta
+          - latest/edge
+        default: latest/stable
+  # Nightly test of latest/stable and latest/beta
+  schedule:
+    - cron: '17 5 * * *'
```
dariuszd21, line 40: nit: maybe we don't have to run it during the weekends also ?

dariuszd21 (changes requested):
Thanks for the PR, only minors 😃

.github/workflows/smoke-test.yml
lczyk: this input picks the channel of the `snapd` snap, not docker

.github/workflows/smoke-test.yml
lczyk: https://github.com/canonical/docker-snap/pull/392/commits/5364d533f50b6bf1dc8bfd9091b01f00e3b7a65b

.github/workflows/smoke-test.yml
lczyk: https://github.com/canonical/docker-snap/pull/392/commits/c0b3077eaf5d561dfce7b09fbb5f7b25b2659dee

.github/workflows/smoke-test.yml
dariuszd21: Right, I think I've messed this up with the other PR 😅

dariuszd21 (approved):
Thanks!

I've re-run failing test as it seemed to be failing due to dockerhub issues

PR #398, opened 2026-09-07
chore: update moby to v29.8.0

### LINKED ISSUE #396: latest/edge spread test failures
Just a quick issue to track the fact that `latest/edge` after #390 is failing the general spread tests due to https://github.com/moby/go-archive/issues/98, which makes it unsuitable for release. we can patch it here or wait for upstream moby to patch to patched `go-archive` depending on urgency

note that #390 *did* pass the ci, so this is also related to https://github.com/canonical/docker-snap/issues/397

dariuszd21:
Update the packages to their latest compatible versions:
- moby `v29.8.0`,
- containerd  `v2.3.4`,
- buildx `v0.37.0`
- compose `v5.5.1`

Patch was hand-modified, will be regenerated before the PR goes out of draft

Fixes: #396


dariuszd21:
I believe spread tests would pass if our `core22` and ` core24` systems updated the `snapd` inside (in my assessment it seems that the problem is with accessing `/proc/mountinfo` now).

lczyk:
@dariuszd21 yeah, agreed. in the meantime we could refresh snapd manually in core guests? it feels not quite right to me though, unless we know that core is aware they need to rebuild their images first. otherwise we're silently accommodating for smth which *would* hurst users who don't refresh. 

dropping analysis from me for reference:

moby 29.x `docker cp` now uses the `fchmodat2` syscall. current snapd allows it in the seccomp template, and the cloud guests get snapd 2.76.3 via `apt install snapd`. ubuntu core guests keep whatever snapd is seeded in the stable core-22/24 images, and image-garden's images hold refreshes *on purpose*. we get older snapd in these images and:

```
$ sudo docker cp <id>:/home/file /var/snap/docker/common/
fchmodat2 file: operation not permitted
```

dariuszd21:
Yeah, but I believe this message is misleading, because it's not `fchmodat2` which is failing but `/proc/mountinfo` :sweat_smile: 

The difference is that in `29.6.0` it wasn't breaking the `cp`, while in `29.7.0` and later it does. As I've mentioned refreshing the `snapd` fixes the problem, the seeded ones are failing on both core22 and core24

```
Sep 08 13:06:01 core22-docker systemd[1]: Started snap.docker.docker.33c8e8ca-f347-411a-8242-6c44d07ed551.scope.
Sep 08 13:06:01 core22-docker audit[2962]: AVC apparmor="DENIED" operation="open" profile="snap.docker.docker" name="/proc/2962/mountinfo" pid=2962 comm="docker" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
Sep 08 13:06:01 core22-docker kernel: audit: type=1400 audit(1788872761.870:50): apparmor="DENIED" operation="open" profile="snap.docker.docker" name="/proc/2962/mountinfo" pid=2962 comm="docker" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
```

lczyk:
@dariuszd21 i've patched up my spread improvements PR and it passes there ( with the reset added ) https://github.com/canonical/docker-snap/actions/runs/34244609329 👍 

edit: amd64 only since that's the only one we get in branch channels

lczyk (dismissed):
lgtm, especially with the spread tests passing

edit: but yeah, pls regenerate the patch

lczyk (approved):

